PR #1, opened 2020-02-25
Get company by id, get branch by id, get all branches by companyId

Raibo:
GET methods for Company and Branch, also refactoring in Update Company

### DIFF
--- a/Api/Services/Customers/CompanyService.cs
+++ b/Api/Services/Customers/CompanyService.cs
@@ -80,49 +80,61 @@ namespace HappyTravel.Edo.Api.Services.Customers
         }
 
 
-        public async Task<Result<CompanyInfo>> Update(CompanyInfo company, int companyId)
+        public Task<Result<CompanyInfo>> Get(int companyId)
         {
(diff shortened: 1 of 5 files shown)


Api/Services/Customers/CompanyService.cs
```diff
@@ -276,6 +316,21 @@ async Task<Result<Company>> GetCompany()
         }
 
 
+        private async Task<Result<Company>> IsCustomerInCompany(int companyId)
```
mingazhev: We use "Is" in method start when method returns Boolean result. This method returns Result, better name for that will be something like 'Check...' or using _.Ensure_ method.

Api/Services/Customers/CompanyService.cs
```diff
@@ -276,6 +316,21 @@ async Task<Result<Company>> GetCompany()
...
+        {
+            var (_, customerCompanyId, _, _) = await _customerContext.GetCustomer();
+
+            var company = await _context.Companies.SingleOrDefaultAsync(c => c.Id == companyId);
```
mingazhev: This logic does not correspond to method's name

PR #3, opened 2020-02-25
Removed verified company check for user invitation

mingazhev:


Api/Controllers/CustomersController.cs
```diff
@@ -103,7 +103,6 @@ public async Task<IActionResult> RegisterInvitedCustomer([FromBody] RegisterInvi
         [HttpPost("customers/invitations")]
         [ProducesResponseType((int) HttpStatusCode.NoContent)]
         [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
-        [MinCompanyState(CompanyStates.ReadOnly)]
```
kirillta: This behavior is correct

PR #5, opened 2020-02-27
Bugfix/nijo 432 fix authorization pipeline

mingazhev:
1. Fixed authorization pipeline. By default, authorization policies in ASP .NET Core are executed even for not authenticated users, which is not suitable for us. I've replaced policy evaluator to change this behaviour.
2. Added new authorize attribute - [CustomerRequired] which should check that current user is bound to customer. This brings us to ability of getting customer without checking a result to failure. We should be sure that customer exists and found.

### DIFF
--- a/Api/Startup.cs
+++ b/Api/Startup.cs
@@ -13,6 +13,7 @@ using HappyTravel.Edo.Api.Filters;
 using HappyTravel.Edo.Api.Filters.Authorization;
 using HappyTravel.Edo.Api.Filters.Authorization.AdministratorFilters;
 using HappyTravel.Edo.Api.Filters.Authorization.CompanyStatesFilters;
+using HappyTravel.Edo.Api.Filters.Authorization.CustomerExistingFilters;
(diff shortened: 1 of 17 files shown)


Api/Startup.cs
```diff
@@ -395,6 +398,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IAuthorizationHandler, InCompanyPermissionAuthorizationHandler>();
             services.AddTransient<IAuthorizationHandler, MinCompanyStateAuthorizationHandler>();
             services.AddTransient<IAuthorizationHandler, AdministratorPermissionsAuthorizationHandler>();
+            services.AddTransient<IAuthorizationHandler, CustomerRequiredAuthorizationHandler>();
+
+            services.AddTransient<IPolicyEvaluator, ForbidUnauthenticatedPolicyEvaluator>();
+            services.AddTransient<PolicyEvaluator>();
```
kirillta: There two lines require clarification

mingazhev: Added comments

PR #7, opened 2020-03-03
Feature/nijo 382 customers search and edit

Raibo:
Added endpoints:
get all customers of a branch, of a company, 
get a single customer,
update permissions of a customer


Api/Controllers/CustomersController.cs
```diff
@@ -157,6 +160,98 @@ public async Task<IActionResult> GetCurrentCustomer()
         }
 
 
+        /// <summary>
+        ///     Get all customers of a company
+        /// </summary>
+        [HttpGet("companies/{companyId}/customers")]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
```
mingazhev: This swagger doc is wrong. There must be a type of response. List of customers?

Api/Controllers/CustomersController.cs
```diff
@@ -157,6 +160,98 @@ public async Task<IActionResult> GetCurrentCustomer()
         }
 
 
+        /// <summary>
+        ///     Get all customers of a company
```
mingazhev: Get**s** (what this method does) , there and further

Api/Models/Customers/CustomerInfoInSearch.cs
```diff
@@ -0,0 +1,31 @@
+using HappyTravel.Edo.Api.Models.Markups;
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Models.Customers
+{
+    public readonly struct CustomerInfoInSearch
```
mingazhev: Is this a public available model? If yes it's properties should be documented

mingazhev: Name looks a little weird I think. Maybe something with using _slim, light_ in name?

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+        {
+            None,
+            Branch,
+            Company
+        }
+
+        public async Task<Result<List<CustomerInfoInSearch>>> GetCustomers(int companyId, int branchId = default)
+        {
+            var customer = await _customerContext.GetCustomer();
+            var (_, failure, error) = ChechCompanyAndBranch(customer, companyId, branchId);
```
mingazhev: There is a typo there

HappyTravel.Edo.Common/Enums/InCompanyPermissions.cs
```diff
@@ -10,17 +10,23 @@ public enum InCompanyPermissions
...
-            PermissionManagement | 
+            PermissionManagementInCompany | 
             CustomerInvitation | 
             AccommodationAvailabilitySearch | 
             AccommodationBooking |
-            ViewCompanyAllPaymentHistory // 126
+            ViewCompanyAllPaymentHistory |
+            PermissionManagementInBranch |
+            ObserveMarkupInCompany |
+            ObserveMarkupInBranch // 1022
```
mingazhev: Does these levels has any meaning for "InCompanyPermissions" enum?

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+            if (failure)
+                return Result.Fail<List<CustomerInfoInSearch>>(error);
+
+            var markupObserveLevel = MarkupObserveLevel.None;
+            if (customer.InCompanyPermissions.HasFlag(InCompanyPermissions.ObserveMarkupInBranch))
+                markupObserveLevel = MarkupObserveLevel.Branch;
+            if (customer.InCompanyPermissions.HasFlag(InCompanyPermissions.ObserveMarkupInCompany))
+                markupObserveLevel = MarkupObserveLevel.Company;
+
+            var query = from cr in _context.CustomerCompanyRelations
```
mingazhev: These short variable names are hard to understand, cause query is long. I suggest to replace them with more meaningful - cr => relation, cu => customer or something similar

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+                join co in _context.Companies
+                    on cr.CompanyId equals co.Id
+                join br in _context.Branches
+                    on cr.BranchId equals br.Id
+                join mp in _context.MarkupPolicies
+                    on cr.CustomerId equals mp.CustomerId into mpn from mp in mpn.DefaultIfEmpty()
+                where branchId == default ? cr.CompanyId == companyId : cr.BranchId == branchId
+                select new {cr, cu, co, br, mp};
+
+            var data = await query.ToListAsync();
```
mingazhev: This variable looks useless - it's name does not tell anything and it can be removed.

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+
+            var results = data.Select(o => 
+                new CustomerInfoInSearch(o.cu.Id, o.cu.FirstName, o.cu.LastName, o.co.Id, o.co.Name, o.br.Id, o.br.Title,
+                    GetMarkupIfPermission(o.mp, o.cr)))
+                .ToList();
+
+            return Result.Ok(results);
+
+            MarkupPolicySettings? GetMarkupIfPermission(MarkupPolicy policy, CustomerCompanyRelation relation)
```
mingazhev: just "GetMarkup" would be enough

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+
+                return null;
+            }
+        }
+
+
+        public async Task<Result<CustomerInfo>> GetCustomer(int companyId, int branchId, int customerId)
+        {
+            var customer = await _customerContext.GetCustomer();
+            var (_, failure, error) = ChechCompanyAndBranch(customer, companyId, branchId);
```
mingazhev: "isFailure"

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+
+            if (foundCustomer == null)
+                return Result.Fail<CustomerInfo>("Customer not found in specified company or branch");
+
+            return Result.Ok(foundCustomer.Value);
+        }
+
+
+        public async Task<Result<List<InCompanyPermissions>>> UpdateCustomerPermissions(int companyId, int branchId, int customerId, 
+            List<InCompanyPermissions> newPermissionsList)
```
mingazhev: "permissions" will be enough I think

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,12 +60,143 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+        {
+            var customer = await _customerContext.GetCustomer();
```
mingazhev: Whole method is too long and hardly readable. Consider splitting it to smaller pieces

HappyTravel.Edo.Common/Enums/InCompanyPermissions.cs
Raibo: Difference between same permissions in a company and in a branch speaks for itself.
In one case customer has permission against all the customers within a company, while in the other case only against these in his own branch.

Api/Models/Customers/CustomerInfoSlim.cs
```diff
@@ -0,0 +1,62 @@
+using HappyTravel.Edo.Api.Models.Markups;
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Models.Customers
+{
+    public readonly struct CustomerInfoSlim
```
kirillta: SlimCustomerInfo

Api/Services/Customers/CustomerPermissionManagementService.cs
```diff
@@ -21,23 +22,41 @@ public class CustomerPermissionManagementService : ICustomerPermissionManagement
...
+                .OnSuccess(GetRelation)
                 .Ensure(IsPermissionManagementRightNotLost, "Cannot revoke last permission management rights")
                 .OnSuccess(UpdatePermissions);
 
-            async Task<Result<CustomerInfo>> GetCurrentCustomer() => await _customerContext.GetCustomerInfo();
+            Result CheckPermission()
+            {
+                if (!customer.InCompanyPermissions.HasFlag(InCompanyPermissions.PermissionManagementInBranch)
+                    && !customer.InCompanyPermissions.HasFlag(InCompanyPermissions.PermissionManagementInCompany))
+                    return Result.Fail("Permission to update customers permissions denied");
```
kirillta: A permission to permissions, awkward sentence

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,11 +59,94 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+
+            var results = (await query.ToListAsync()).Select(o => 
+                new CustomerInfoSlim(o.customer.Id, o.customer.FirstName, o.customer.LastName,
+                    o.company.Id, o.company.Name, o.branch.Id, o.branch.Title,
+                    GetMarkup(o.policy, o.relation)))
+                .ToList();
+
+            return Result.Ok(results);
+
+            MarkupPolicySettings? GetMarkup(MarkupPolicy policy, CustomerCompanyRelation relation)
```
kirillta: MarkupPolicySettings must not be null

kirillta: I suggest to use some kind of an empty constructor

kirillta:
I expect you'll also add tests to this request

Api/Services/Customers/CustomerPermissionManagementService.cs
Raibo: I don't really know how to make this less awkward, any ideas?

Api/Models/Customers/CustomerCompanyInfo.cs
```diff
@@ -24,6 +26,16 @@ public CustomerCompanyInfo(int id, string name, bool isMaster, List<InCompanyPer
...
 
+        /// <summary>
+        ///     Id of the branch of the company, to which the customer belongs.
+        /// </summary>
+        public int BranchId { get; }
+
+        /// <summary>
+        ///     Title of the branch.
+        /// </summary>
+        public string BranchTitle { get; }
```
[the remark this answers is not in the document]
Raibo: That's how it really is, entity 'Branch' actually have 'Title', not 'Name' property.

Api/Services/Customers/CustomerPermissionManagementService.cs
kirillta: "You have no acceptance to manage customers permissions"

Api/Models/Customers/CustomerCompanyInfo.cs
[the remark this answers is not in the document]
kirillta: Then it's in your hands to fix that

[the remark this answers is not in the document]
Raibo: This is a rather big change, I suggest doing that in a separate pr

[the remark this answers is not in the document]
Raibo: Changed in DTO classes tho, so that frontend could use branchName fields

Api/Services/Customers/CustomerService.cs
Raibo: There was a long conversation outside github, so that this field is string now. 
In this case its empty string

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,11 +61,105 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+            var customerIdList = relations.Select(x => x.customer.Id).ToList();
+
+            var markupsMap = (await (
+                from markup in _context.MarkupPolicies
+                where markup.CustomerId != null 
+                    && customerIdList.Contains(markup.CustomerId.Value)
+                    && markup.ScopeType == MarkupPolicyScopeType.Customer
+                select markup)
+                .ToListAsync())
+                .GroupBy(k => (int)k.CustomerId)
```
kirillta: May it be grouped on a database?

Api/Services/Customers/CustomerService.cs
```diff
@@ -54,11 +61,105 @@ public async Task<Result<Customer>> GetMasterCustomer(int companyId)
...
+                    o.customer.Created, o.company.Id, o.company.Name, o.branch.Id, o.branch.Title,
+                    GetMarkup(o.relation)))
+                .ToList();
+
+            return Result.Ok(results);
+
+            string GetMarkup(CustomerCompanyRelation relation)
+            {
+                if (!markupsMap.TryGetValue(relation.CustomerId, out var policies))
+                    return "";
```
kirillta: Replace it to string.Empty

Api/Services/Customers/CustomerService.cs
Raibo: I doubt. We can't use aggregated results like 'sum' or 'average' here, we need all rows for each group. Getting individual query for every Id would be a bad option as well.

PR #9, opened 2020-03-05
Feature/nijo 440 including currency converter

mingazhev:
In this PR:
- Removed old CurrencyRate service and it's artifacts, replaced with new implementation
- Added currency conversion to show prices for 1,2,3 steps of availability search based on Customer preferred currency
- Refactored markup application and shortened replaced some weird models with generic


Api/Services/Accommodations/AvailabilityService.cs
```diff
@@ -125,58 +140,70 @@ ProviderData<SingleAccommodationAvailabilityDetails> AddProviderData(SingleAccom
...
+            var (_, _, settings, _) = await _customerSettingsManager.GetUserSettings(customer);
+            var preferredCurrency = settings.PreferredCurrency;
+
+            if (preferredCurrency == default)
+                return details;
+
+            return await func(details, async (price, currency) =>
+            {
+                var newCurrency = preferredCurrency;
+                var newPrice = price * await _currencyRateService.Get(currency, preferredCurrency);
```
kirillta: This conversion is pretty naive

mingazhev: What do you mean?
It works incorrect, slowly?

mingazhev: If it's about no error handling, you're right, fixing that now

kirillta: Not only that — ceiling, sane checks and other.
Should we make a separate library for that?

PR #13, opened 2020-03-10
Bugfix/nijo 458 fix data serialization

mingazhev:
1. We store serialized `BookingDetails ` (from contracts) along with booking in JSONB column in DB.
I've noticed that in some places we deserialize it to another model, which looks quite similar:
`AccommodationBookingDetails`.

It worked in most cases (types look similar) but it's not very safe. I've changed that to deserialize to valid type. 

2. Moved provider interaction methods from `BookingManager `to `BookingService`.
Idea is that `BookingManager `is responsible for storing and updating bookings and `BookingService `is the main service with common logic, which uses BookingManager for it's needs.

3. Moved booking payments logic to `BookingPaymentService` from `PaymentService`


Api/Services/Accommodations/Bookings/BookingService.cs
```diff
@@ -93,7 +104,47 @@ public class BookingService : IBookingService
...
+                }
+                catch (Exception ex)
+                {
+                    var errorMessage = $"Failed to update booking data (refcode '{referenceCode}') after the request to the connector";
+
+                    var (_, isCancellationFailed, cancellationError) = await _providerRouter.CancelBooking(booking.DataProvider, booking.ReferenceCode);
+                    if (isCancellationFailed)
+                        errorMessage += Environment.NewLine + $"Booking cancellation has failed: {cancellationError}";
+
+                    _logger.LogError(ex, errorMessage);
```
kirillta: Use logger delegate here

kirillta: You added delegates, but forgot to replace this very line

Api/Services/Accommodations/Bookings/BookingPaymentService.cs
```diff
@@ -0,0 +1,335 @@
...
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Services.Accommodations.Bookings
+{
+    public class BookingPaymentService : IBookingPaymentService
+    {
+        public async Task<Result<List<int>>> GetBookingsForCapture(DateTime deadlineDate)
+        {
+            if (deadlineDate == default)
```
kirillta: We have nullable deadlines now

mingazhev: This is not a problem there. This 'deadlineDate' is a parametr, which is sent from controller to get list of booking id's.
Anyway, I've added capturing money after checkin date too.

Api/Services/Accommodations/Bookings/BookingService.cs
mingazhev: fixed

PR #27, opened 2020-03-27
Changed credit card models and controller endpoints

mingazhev:
1. Added separate endpoint to pay with saved card (to fix the bug)
2. Little refactoring of creditCardsPaymentService - preparing to move part of its logic to separate service

### DIFF
--- a/Api/Controllers/CreditCardsController.cs
+++ b/Api/Controllers/CreditCardsController.cs
@@ -39,8 +39,8 @@ namespace HappyTravel.Edo.Api.Controllers
         [CustomerRequired]
         public async Task<IActionResult> Get()
         {
-            var customerInfo = await _customerContext.GetCustomer();
-            return Ok(await _cardService.Get(customerInfo));
+            var customer = await _customerContext.GetCustomer();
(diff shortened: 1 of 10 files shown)


Api/Controllers/CreditCardsController.cs
```diff
@@ -33,45 +36,24 @@ public CreditCardsController(ICreditCardService cardService, ICustomerContext cu
...
         [ProducesResponseType(typeof(CreditCardInfo[]), (int) HttpStatusCode.OK)]
         [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
+        [CustomerRequired]
         public async Task<IActionResult> Get()
         {
-            var (_, isFailure, customerInfo, error) = await _customerContext.GetCustomerInfo();
-            if (isFailure)
-                return BadRequest(ProblemDetailsBuilder.Build(error));
-
+            var customerInfo = await _customerContext.GetCustomer();
```
hasmikharutyunyan: maybe customer?

mingazhev: fixed

PR #29, opened 2020-03-30
Master's email address added to the registration message

kirillta:


kirillta:
I'll merge this one, because it has no significant changes

PR #33, opened 2020-04-03
Update the contracts up to 1.2.5

oleg1std:


oleg1std:
Do not merge!

Api/Models/Accommodations/AvailabilityResult.cs
```diff
@@ -30,6 +30,6 @@ public AvailabilityResult(AvailabilityResult result, List<Agreement> agreements)
         /// <summary>
         /// List of available agreements
         /// </summary>
-        public List<Agreement> Agreements { get; }
+        public List<RoomContractSet> Agreements { get; }
```
hasmikharutyunyan, line 33: Why not RoomContractSets?

oleg1std: It's an external model and changing property name can break something on the web-app

kirillta: That will be done sooner or later. Think it's safe to rename now

Api/Services/Connectors/DataProvider.cs
```diff
@@ -40,21 +40,21 @@ public DataProvider(IDataProviderClient dataProviderClient, string baseUrl, ILog
...
         }
 
 
-        public Task<Result<DeadlineDetails, ProblemDetails>> GetDeadline(string availabilityId, Guid agreementId, string languageCode)
+        public Task<Result<DeadlineDetails, ProblemDetails>> GetDeadline(string availabilityId, Guid roomContractSetId, string languageCode)
         {
             return ExecuteWithLogging(() =>
             {
-                var uri = new Uri($"{_baseUrl}accommodations/availabilities/{availabilityId}/agreements/{agreementId}/deadline", UriKind.Absolute);
+                var uri = new Uri($"{_baseUrl}accommodations/availabilities/{availabilityId}/room-contract-sets/{roomContractSetId}/deadline", UriKind.Absolute);
```
kirillta, line 57: Don't forget to change routes in Insomnia

PR #40, opened 2020-04-08
Added money rounding for availability response

mingazhev:
1. Prices from connectors sometimes can come with lots of digits (can be 4 digits for Illusions)
2. Prices that we process during currency conversion and applying markups can have many digits too. 

In this PR I've added money rounding right before returning models from availability service - at markups applying step.


kirillta:
Money must be only ceiled. Also we have similar function in another project

PR #42, opened 2020-04-09
Feature/nijo 498 add etg webhook

oleg1std:
The Pr adds an Etg's webhook.

DO NOT MERGE!


Api/Controllers/BookingResponseController.cs
```diff
@@ -45,6 +49,18 @@ public async Task<IActionResult> HandleNetstormingBookingResponse()
...
 
 
+        [AllowAnonymous]
+        [ProducesResponseType((int) HttpStatusCode.OK)]
+        [ProducesResponseType((int) HttpStatusCode.BadRequest)]
+        [HttpPost("bookings/accommodations/responses/etg")]
+        public async Task<IActionResult> HandleEtgBookingResponse(EtgBookingResponse etgBookingResponse)
+        {
+            var (_, isFailure, error) = await _etgResponseService.ProcessBookingStatus(etgBookingResponse);
+            return Ok(isFailure ? error : "ok");
```
mingazhev, line 59: This is what connector waits from our service?
What does "ok" mean?

Api/Startup.cs
```diff
@@ -536,6 +547,24 @@ private IAsyncPolicy<HttpResponseMessage> GetDefaultRetryPolicy()
...
+        private void AddEtgOptionsFromVault(IServiceCollection services)
+        {
+            using var vaultClient = new VaultClient.VaultClient(new VaultOptions
+            {
+                BaseUrl = new Uri(EnvironmentVariableHelper.Get("Vault:Endpoint", Configuration)),
+                Engine = Configuration["Vault:Engine"],
+                Role = Configuration["Vault:EtgRole"]
+            });
+            vaultClient.Login(EnvironmentVariableHelper.Get("Vault:Token", Configuration)).Wait();
+            var values = vaultClient.Get(Configuration["Etg:Options"]).Result;
```
mingazhev, line 560: Isn't this code duplicates another one?
I think vault client should be created once, and login procedure should execute once during startup

Api/Services/ProviderResponses/EtgResponseService.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (isGetBookingDetailsFailure)
+                return Result.Fail(getBookingDetailsError.Detail);
+                
+            await _customerContext.SetCustomerInfo(booking.CustomerId);
+            
+            return await _bookingService.ProcessResponse(bookingDetails, booking); 
+        }
+
+
+        private bool CheckSignature(EtgBookingResponseSignature responseSignature)
```
mingazhev, line 52: In my opinion this logic is too provider-specific to place it in EDO

PR #47, opened 2020-04-15
Added credit card payments validation

mingazhev:
Although cardholder name is not strictly required for all kind of payments it must be filled if customer wants to save card for further payments.
UI validation should be done on front end side, in this PR I've added back end validation.

### DIFF
--- a/Api/Services/Payments/CreditCards/CreditCardPaymentService.cs
+++ b/Api/Services/Payments/CreditCards/CreditCardPaymentService.cs
@@ -2,10 +2,12 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
+using FluentValidation;
 using HappyTravel.Edo.Api.Infrastructure;
 using HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions;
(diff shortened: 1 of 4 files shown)


Api/Services/Payments/CreditCards/CreditCardPaymentService.cs
```diff
@@ -49,10 +51,30 @@ public class CreditCardPaymentProcessingService : ICreditCardPaymentProcessingSe
...
                 .OnSuccessIf(IsSaveCardNeeded, SaveCard)
                 .OnSuccess(StorePaymentResults);
 
+
+            static Result Validate(NewCreditCardPaymentRequest payment)
+            {
+                return GenericValidator<NewCreditCardPaymentRequest>
+                    .Validate(p =>
+                    {
+                        if (payment.IsSaveCardNeeded)
```
kirillta, line 65: Shouldn't we validate other fields?

mingazhev: I've added attributes for other required fields in model.
Looks not very obvious, sure.
Should I move validation to here?

kirillta: No

PR #50, opened 2020-04-22
Updated EDO to use contracts v1.3.1

mingazhev:


Api/Models/Bookings/SlimRoomContract.cs
```diff
@@ -9,17 +9,13 @@ namespace HappyTravel.Edo.Api.Models.Bookings
         public SlimRoomContract(RoomContract roomContract)
         {
             MealPlan = roomContract.MealPlan;
-            MealPlanCode = roomContract.MealPlanCode;
-            ContractType = roomContract.ContractType;
-            BoardBasisCode = roomContract.BoardBasisCode;
-            BoardBasis = roomContract.BoardBasis;
+            ContractType = roomContract.ContractDescription;
+            BoardBasis = roomContract.BoardBasis.ToString();
```
hasmikharutyunyan, line 13: Why we don't use enum here?

mingazhev: fixed

PR #55, opened 2020-04-27
On demand updating booking details from connector

mingazhev:


Api/Services/Accommodations/Bookings/IBookingManager.cs
```diff
@@ -24,11 +24,11 @@ public interface IBookingManager
         
         Task<Result<List<SlimAccommodationBookingInfo>>> GetAgentBookingsInfo();
         
-        Task<Result> ConfirmBooking(BookingDetails bookingDetails, Booking booking);
+        Task ConfirmBooking(BookingDetails bookingDetails, Booking booking);
```
kirillta, line 27: Again too many "Bookings" around

Api/Services/Accommodations/Bookings/BookingService.cs
```diff
@@ -287,6 +273,22 @@ async Task<Result> LogAppliedMarkups()
         }
 
 
+        public async Task<Result<BookingDetails, ProblemDetails>> Refresh(int bookingId)
```
kirillta, line 276: The method makes sense only for sync connectors

Api/Controllers/AccommodationsController.cs
```diff
@@ -191,6 +191,26 @@ public async Task<IActionResult> FinalizeBooking([FromRoute] string referenceCod
             return Ok(bookingDetails);
         }
 
+
+        /// <summary>
+        ///     Sends booking request to a data provider to get refreshed booking details, especially - status.
+        /// </summary>
+        /// <param name="bookingId">Id of the booking</param>
+        /// <returns>Updated booking details.</returns>
+        [HttpPost("accommodations/bookings/{bookingId}/refresh")]
```
kirillta, line 200: 1. Should be `check-status` or something from user's perspective
2. Definitely this is GET

mingazhev: This method actually updates database entry and is not idempotent at all. Why do you think it is "GET"

Api/Services/Accommodations/Bookings/IBookingManager.cs
mingazhev: fixed

Api/Services/Accommodations/Bookings/BookingService.cs
mingazhev: In my thought It should be supported for all connectors, by contract.
To differ one from another I'll add a flag to 'Booking' entity, in further PRs:
`IsAsync` or maybe some kind of enum: `BookingUpdateMode: { Synchronous, Asynchronous }`

mingazhev: EDO could use this flag to:
1. Return to frontend where it could show the refresh button or not
2. Query bookings that need to be updated from supplier
3. Execute provider request to refresh booking or not

Api/Controllers/AccommodationsController.cs
mingazhev: The semantic of this method is: Get data from somewhere and update and I think we should not hide this from user

kirillta: It may update or may not. Update is a secondary function and executes under the hood. At least it should called `refresh-status`

mingazhev: changed to 'refresh-status'

mingazhev: changed to `refresh-status`

Api/Services/Accommodations/Bookings/IBookingManager.cs
kirillta: You changed only one method of three

mingazhev: Will fix that in further PR

PR #58, opened 2020-05-01
Add location normalizer service

oleg1std:


Api/Controllers/LocationsController.cs
```diff
@@ -105,7 +106,17 @@ public async Task<IActionResult> GetLastModifiedDate()
         }
 
 
+        [ProducesResponseType(typeof(DateTime), (int) HttpStatusCode.OK)]
```
mingazhev, line 109: This is not true, nor the response type nor the status code

Api/Infrastructure/ServiceCollectionExtensions.cs
```diff
@@ -403,6 +404,9 @@ public static IServiceCollection AddServices(this IServiceCollection services)
 
             services.AddTransient<INetstormingResponseService, NetstormingResponseService>();
             services.AddTransient<IBookingWebhookResponseService, BookingWebhookResponseService>();
+
+            services.AddNameNormalizationServices();
+            services.AddSingleton<ILocationNormalizer, LocationNormalizer>();
```
mingazhev, line 409: It depends on scoped service - EdoContext. Does this work?

Api/Services/Locations/LocationNormalizer.cs
```diff
@@ -0,0 +1,120 @@
...
+        private void Normalize(List<Location> locations)
+        {
+            foreach (var location in locations)
+            {
+                Normalize(location);
+            }
+        }
+
+
+        void NormalizeCountry(Location location)
```
mingazhev, line 55: Need to add access modifiers there and further, or move to local functions

Api/Infrastructure/Logging/LoggerEvents.cs
```diff
@@ -32,5 +32,6 @@ public enum LoggerEvents
         AgentAuthorizationFailure = 1111,
         CounterpartyStateAuthorizationSuccess = 1120,
         CounterpartyStateAuthorizationFailure = 1121,
+        LocationNormalized = 1122
```
mingazhev, line 35: I suggest to add a spacing between logger events which are far from each other, e.g. 1130.
In this case we could add new event ids right after 'CounterpartyStateAuthorizationFailure'

Api/Controllers/LocationsController.cs
mingazhev: And method should be documented

Api/Infrastructure/ServiceCollectionExtensions.cs
oleg1std: Yes, it works. But, I agree, it is a mismatch,

PR #66, opened 2020-05-07
Feature/nijo 332 remove booking details

mingazhev:


HappyTravel.Edo.Data/Booking/LocationInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+        {
+            CountryCode = countryCode;
+            CountryName = countryName;
+            CityCode = cityCode;
+            CityName = cityName;
+        }
+        
+        public string CountryCode { get; }
+        public string CountryName { get; }
+        public string CityCode { get; }
```
kirillta, line 15: It's a locality code. And do we need codes here?

kirillta: Also, where's a zone name?

HappyTravel.Edo.Data/EdoContext.cs
```diff
@@ -527,6 +524,26 @@ private void BuildBooking(ModelBuilder builder)
                 booking.Property(b => b.LanguageCode)
                     .IsRequired()
                     .HasDefaultValue("en");
+                
+                booking.Property(b => b.AccommodationId)
+                    .IsRequired();
+                
+                booking.Property(b => b.AccommodationName)
```
kirillta, line 531: Do we need the name as a separate column?

mingazhev: Thought that if we have an id as separate column, why not to do the same with name?

HappyTravel.Edo.Data/Booking/LocationInfo.cs
mingazhev: Removed codes, renamed accordingly

PR #75, opened 2020-05-11
Invoice data filling

mingazhev:


Api/Models/Mailing/BookingInvoiceData.cs
```diff
@@ -67,5 +65,24 @@ public SellerInfo(string companyName, string bankName, string bankAddress, strin
             public string RoutingCode { get; }
             public string SwiftCode { get; }
         }
+        
+        
+        public readonly struct InvoiceItem
```
kirillta, line 70: This one should name `InvoiceInfo` according to adjacent models

Api/Models/Mailing/BookingVoucherData.cs
```diff
@@ -25,10 +24,11 @@ namespace HappyTravel.Edo.Api.Models.Mailing
 
 
         public int BookingId { get; }
+        public string AgentName { get; }
         public AccommodationInfo Accommodation {get;}
-        public DateTime CheckInDate { get; }
-        public DateTime CheckOutDate { get; }
-        public DateTime? DeadlineDate { get; }
+        public string CheckInDate { get; }
```
kirillta, line 29: Better to separate mail and front-end models, because we limited in a server-side formatting.

Api/Models/Mailing/BookingInvoiceData.cs
mingazhev: InvoiceItemInfo maybe? This is not about the whole invoice, but about a row in a table

Api/Models/Mailing/BookingVoucherData.cs
mingazhev: fixed

PR #84, opened 2020-05-14
Adding request metadata to method params

mingazhev:
We use request ids to trace and log requests through all our services.
In previous version it was generated inside a dataprovider client with using HttpContextAccessor.
In new version there is a new model: RequestMetadata, which contains language code and request id and should be passed to methods explicitly.
This allows us to use this functionality in async use cases.


kirillta:
Doubt we need it _that_ way

mingazhev:
This is not actual since we decided to use OpenTelemetry approach

PR #85, opened 2020-05-15
Async availability search

mingazhev:
There are three new endpoints:
1. Starts a search and returns an id
2. Gets the **state** of previous started search (by given id)
3. Gets the **result** of previous started search (by given id)

+ some code was temproary commented out. This is the first implementation version and further it will be reimplemented to not wait untill all connectors will return their result.
For now it works the same as previous implementation


Api/Services/Accommodations/Availability/AvailabilityStorage.cs
```diff
@@ -0,0 +1,52 @@
...
+using FloxDc.CacheFlow;
+using FloxDc.CacheFlow.Extensions;
+using HappyTravel.Edo.Api.Infrastructure.Converters;
+using HappyTravel.Edo.Api.Models.Accommodations;
+
+namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
+{
+    public class AvailabilityStorage 
+    {
+        public AvailabilityStorage(IJsonSerializer serializer,
```
kirillta, line 12: Unused `IJsonSerializer`

Api/Services/Accommodations/Availability/MultiProviderAvailabilitySearchService.cs
```diff
@@ -0,0 +1,99 @@
...
+            _storage = storage;
+            _logger = logger;
+        }
+
+
+        public Guid StartSearch(AvailabilityRequest availabilityRequest, AgentInfo agent, string languageCode)
+        {
+            var searchId = Guid.NewGuid();
+#pragma warning disable 4014 This method should not be awaited and should run in new thread
+            StartSearch(searchId, availabilityRequest, agent, languageCode);
```
kirillta, line 28: Maybe if you run a task and discard its result, the line would look more readable?

Api/Services/Accommodations/Availability/AvailabilityStorage.cs
mingazhev: Fixed

Api/Services/Accommodations/Availability/MultiProviderAvailabilitySearchService.cs
mingazhev: Fixed

PR #88, opened 2020-05-18
Removed prices from voucher data

mingazhev:


Api/Models/Mailing/BookingVoucherData.cs
```diff
@@ -50,5 +50,26 @@ public AccommodationInfo(string name, in SlimLocationInfo locationInfo, in Conta
             public SlimLocationInfo Location { get; }
             public string Name { get; }
         }
+        
+        public readonly struct RoomInfo
```
hasmikharutyunyan, line 54: Maybe will be better to move it to separate file?

mingazhev: I think as soon as this is totally part of this data structure and not used anywhere else, it's a good way to store it in the same file. And so Kirill did, for other similar structures there

PR #97, opened 2020-05-26
added migration for spped up predictions search

hasmikharutyunyan:


HappyTravel.Edo.Data/Migrations/20200526113018_AddedCalculatedColumnForTsVector.cs
```diff
@@ -0,0 +1,58 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Create generated Column for ConcatenatedTsVector
+            migrationBuilder.Sql(
+                @" ALTER TABLE ""Locations"" 
+                    ADD COLUMN ""ConcatenatedTsVector"" tsvector GENERATED ALWAYS AS (get_location_concatenated_tsvector(""Name"",""Locality"",""Country"")) STORED");
+            
+            migrationBuilder.Sql("DROP INDEX \"IDX_GetLocationConcatenatedTsVector\";");
+
+            var searchFunctionSql = @"
```
kirillta, line 16: I would like to see more explanations here, like what `select *` includes, or why that returns four location types?

PR #105, opened 2020-06-02
Invalid bookings cancellation

mingazhev:
1. "All unpaids bookings should be cancelled - with no dependency on deadline or payment method" - this logic was implemented
2. Tests

### DIFF
--- a/Api/Controllers/InternalBookingsController.cs
+++ b/Api/Controllers/InternalBookingsController.cs
@@ -25,16 +25,15 @@ namespace HappyTravel.Edo.Api.Controllers
 
 
         /// <summary>
-        ///     Gets bookings for cancellation by deadline date
+        ///     Gets bookings for cancellation
         /// </summary>
-        /// <param name="deadlineDate">Deadline date</param>
         /// <returns>List of booking ids for cancellation</returns>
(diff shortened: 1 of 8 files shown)


Api/Controllers/InternalBookingsController.cs
```diff
@@ -25,16 +25,15 @@ public class InternalBookingsController : BaseController
 
 
         /// <summary>
-        ///     Gets bookings for cancellation by deadline date
+        ///     Gets bookings for cancellation
         /// </summary>
-        /// <param name="deadlineDate">Deadline date</param>
         /// <returns>List of booking ids for cancellation</returns>
-        [HttpGet("cancel/{deadlineDate}")]
+        [HttpGet("cancel")]
```
hasmikharutyunyan, line 31: maybe `for-cancel` ?

mingazhev: maybe get-for-cancel? to be a verb

hasmikharutyunyan: I think method already marked as `httpget` and no need of `get` in route.

mingazhev: I'll rethink that in the next iteration

PR #116, opened 2020-06-10
Feature/nijo 665 add authentication to connectors[DO NOT MERGE]

hasmikharutyunyan:


Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -53,28 +57,34 @@ public DataProviderClient(IHttpClientFactory clientFactory, ILogger<DataProvider
...
         {
             try
             {
                 using var client = _clientFactory.CreateClient();
-                
+
                 client.DefaultRequestHeaders.Add("Accept-Language", languageCode);
+                var (_, isFailure, token, getTokenError) = await GetToken();
+                if (isFailure)
+                {
```
mingazhev, line 78: I suggest omit curly brackets there

Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -96,9 +106,34 @@ private static StringContent BuildContent<T>(T requestContent)
         }
 
 
+        private async Task<Result<string, ProblemDetails>> GetToken()
+        {
+            await TokenSemaphore.WaitAsync();
+            var now = DateTime.UtcNow;
```
mingazhev, line 112: please, use IDateTimeProvider instead of hard implementation

Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -96,9 +106,34 @@ private static StringContent BuildContent<T>(T requestContent)
...
+            // We need to cache token because users can send several requests in short periods.
+            if (!_tokenInfo.Equals(default) && _tokenInfo.ExpiryDate >= now)
+                return _tokenInfo.Token;
+
+            using var client = _clientFactory.CreateClient(HttpClientNames.Identity);
+           
+            var tokenResponse = await client.RequestClientCredentialsTokenAsync(_tokenRequest);
+            if (tokenResponse.IsError)
```
mingazhev, line 120: This will be more readable with curly brackets

Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -96,9 +106,34 @@ private static StringContent BuildContent<T>(T requestContent)
...
+                Result.Failure<string, ProblemDetails>(
+                    ProblemDetailsBuilder.Build($"Something went wrong while requesting the access token. Error: {tokenResponse.Error}"));
+
+            _tokenInfo = (tokenResponse.AccessToken, now.AddSeconds(tokenResponse.ExpiresIn));
```
mingazhev, line 124: Could there be a situation when token will expire before usage?
For example, token will expire in one millisecond, we think that it is not expired but at the time we'll use it (a bit later) it will expire.

Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -53,28 +57,34 @@ public DataProviderClient(IHttpClientFactory clientFactory, ILogger<DataProvider
...
             try
             {
                 using var client = _clientFactory.CreateClient();
-                
+
                 client.DefaultRequestHeaders.Add("Accept-Language", languageCode);
+                var (_, isFailure, token, getTokenError) = await GetToken();
+                if (isFailure)
+                {
+                    return Result.Failure<TResponse, ProblemDetails>(getTokenError);
```
mingazhev, line 79: This situation should be logged

Api/Infrastructure/DataProviders/DataProviderClient.cs
```diff
@@ -96,9 +106,34 @@ private static StringContent BuildContent<T>(T requestContent)
         }
 
 
+        private async Task<Result<string, ProblemDetails>> GetToken()
```
mingazhev, line 109: Could we use value task there? (I'm not sure about how it will work with `await TokenSemaphore.WaitAsync();`)

Api/Infrastructure/DataProviders/DataProviderClient.cs
hasmikharutyunyan: I think can be that kind of situation and I can change >= to >, or if this will happen often in the future we can check that difference must be >= 5 second (for example)

Api/Infrastructure/DataProviders/DataProviderClient.cs
hasmikharutyunyan: I think no sense, because `TokenSemaphore.WaitAsync()` returns a Task and it is works asynchronously

Api/Infrastructure/DataProviders/DataProviderClient.cs
mingazhev: Yes, I suggest to add a several seconds lag  to be sure that all will work fine

PR #117, opened 2020-06-14
Added receipt service and entity

mingazhev:
1. Created a separate service for documents registration
2. Added receipt depending on an invoice creation
3. Changed a time of invoice creation, now it's created while booking registration
4. Removed multiple sending receipt logic from multiple services


Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
```diff
@@ -151,12 +149,48 @@ static BookingInvoiceData.SellerInfo GetSellerDetails(Booking booking, BankDetai
                 counterparty.Phone,
                 "billingEmail@mail.com");
```
hasmikharutyunyan, line 150: why we hardcoded this email?

Api/Services/Documents/PaymentDocumentsStorage.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public async Task<DocumentRegistrationInfo> Register<TPaymentDocumentEntity>(TPaymentDocumentEntity documentEntity, Func<int, DateTime, string> numberGenerator)
+            where TPaymentDocumentEntity : class, IPaymentDocumentEntity
+        {
+            var now = _dateTimeProvider.UtcNow();
+            documentEntity.Date = now;
+            _context.Add(documentEntity);
+            // Saving entity to fill it's id
+            await _context.SaveChangesAsync();
+            documentEntity.Number = numberGenerator(documentEntity.Id, now);
```
hasmikharutyunyan, line 32: can we generate this in database ?

Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
mingazhev: this will be rewritten in the next PR. For now we do not have a shared vision of where to get it from - from agency or counterparty.

Api/Services/Documents/PaymentDocumentsStorage.cs
mingazhev: Yes we can. But I think this is a domain logic, and it's better to put it in our code instead of writing this logic in database

PR #118, opened 2020-06-16
Feature/nijo 577 rights without counterparty checking

Raibo:
- No dependency on counterparty in permission checks.
- PermissionChecker work again.
- Several checks of request data to be connected to actual agent's agency. (otherwise permission in one agency allows doing things in other agencies in several cases)


Api/Extensions/AgentInfoExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Data;
+using Microsoft.EntityFrameworkCore;
+
+namespace HappyTravel.Edo.Api.Extensions
+{
+    public static class AgentInfoExtensions
```
hasmikharutyunyan, line 11: why It is done in separate class? I think pass database context to extension method it is not good idea

Raibo: similar checks required all across the services.
so where to keep this code?

Raibo: also, why passing db context to extension method is not good?

hasmikharutyunyan: because, i think extension methods must not be dependent from external resources

hasmikharutyunyan: I think, it can be in AgentService and from other services use this

Api/Services/Agents/AgentPermissionManagementService.cs
```diff
@@ -32,32 +30,12 @@ public class AgentPermissionManagementService : IAgentPermissionManagementServic
         {
             var agent = await _agentContext.GetAgent();
 
-            return await CheckPermission()
-                .Bind(CheckCounterpartyAndAgency)
+            return await Result.Ok()
+                .Ensure(() => agent.IsCurrentAgency(agencyId), "You can only edit permissions of agents from your current agency")
```
mingazhev, line 34: [optional] this will be more beatiful and readable if extracted to method

Api/Services/Agents/AgentService.cs
```diff
@@ -124,18 +124,21 @@ string GetMarkupFormula(AgentAgencyRelation relation)
             {
                 if (!markupsMap.TryGetValue(relation.AgentId, out var policies))
                     return string.Empty;
-                
-                if (currentAgent.InAgencyPermissions.HasFlag(InAgencyPermissions.ObserveMarkupInCounterparty)
-                    || currentAgent.InAgencyPermissions.HasFlag(InAgencyPermissions.ObserveMarkupInAgency) && relation.AgencyId == agencyId)
-                    return _markupPolicyTemplateService.GetMarkupsFormula(policies);
 
-                return string.Empty;
+                if (!isObserveMarkupPermission)
```
mingazhev, line 128: maybe "**has**ObserveMarkupPermissions"?

HappyTravel.Edo.Common/Enums/InAgencyPermissions.cs
```diff
@@ -9,24 +9,20 @@ namespace HappyTravel.Edo.Common.Enums
...
             AgentInvitation | 
             AccommodationAvailabilitySearch | 
             AccommodationBooking |
-            ViewCounterpartyAllPaymentHistory |
-            PermissionManagementInAgency |
-            ObserveMarkupInCounterparty |
-            ObserveMarkupInAgency // 1022
+            PermissionManagement |
+            ObserveMarkup |
+            ObserveAgents, // 254
```
mingazhev, line 26: Should you add "ObserveBalance " there?

Api/Extensions/AgentInfoExtensions.cs
mingazhev: I'm agree, passing the whole context and having such a complex logic in extension method looks a bit weird for me.
AgentService looks like a good candidate for checking that agent is affiliated with counterparty or agency

Api/Extensions/AgentInfoExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public static bool IsCurrentAgency(this AgentInfo agentInfo, int agencyId) =>
```
mingazhev, line 13: "Current" word sounds like some context that "implicitly exists here at this time". And should these two methods (IsCurrentCounterparty, IsCurrentAgency) be exctracted at all? 
I couldn't make up  valid and on-the-nose names for them.

Api/Services/Agents/AgentPermissionManagementService.cs
Raibo: the problem is, there are many places with such a check and creating a method for each would be a massive code duplication

Api/Extensions/AgentInfoExtensions.cs
Raibo: moved to IAgentContextService

Api/Services/Agents/IAgentContextService.cs
```diff
@@ -7,12 +7,16 @@
...
     {
-        ValueTask<AgentInfo> GetAgent();
+        ValueTask<AgentContext> GetAgent();
 
         Task<List<AgentAgencyInfo>> GetAgentCounterparties();
         
-        ValueTask<Result<AgentInfo>> SetAgentInfo(int agentId);
+        ValueTask<Result<AgentContext>> SetAgentInfo(int agentId);
+
+        Task<bool> IsAgentAffiliatedWithCounterparty(int agentId, int counterpartyId);
```
mingazhev, line 18: Here could be used current. If we mean that agentContextService is a service, giving information about current agent, we could use "GetCurrent" instead of "GetAgent" and "IsCurrentAgentAffiliatedWithCounterparty" with only one param - "counterpartyId"

PR #129, opened 2020-06-26
Moved sending receipt to step after successful booking

mingazhev:
Additionally - invoice is generated after successfull booking only


Api/Services/Accommodations/Bookings/BookingService.cs
```diff
@@ -141,9 +145,23 @@ public class BookingService : IBookingService
             }
 
 
+            Task ProcessResponse(BookingDetails bookingResponse) => this.ProcessResponse(bookingResponse, booking);
```
Raibo, line 148: On one hand this looks prettier, on the other hand, you have to look for one more place to get to the actual method logic. I suggest lambda, as it was.

mingazhev: I think this lambda has no important information to show it and all other methods are written as shortened method groups. So I'd leave it how it is done

PR #132, opened 2020-07-01
Add ParentId and remove IsDefault from Agency model

Raibo:
Remove IsDefault
Add ParentId
Comment out AddAgency method


Api/Controllers/CounterpartiesController.cs
```diff
@@ -22,24 +22,24 @@ public CounterpartiesController(ICounterpartyService counterpartyService)
...
-        [AgentRequired]
-        public async Task<IActionResult> AddAgency(int counterpartyId, [FromBody] AgencyInfo agencyInfo)
-        {
-            var (isSuccess, _, _, error) = await _counterpartyService.AddAgency(counterpartyId, agencyInfo);
+        ///// <summary>
+        /////     Creates agency for counterparty.
+        ///// </summary>
+        ///// <param name="counterpartyId">Counterparty Id.</param>
+        ///// <param name="agencyInfo">Agency information.</param>
+        ///// <returns></returns>
```
mingazhev, line 30: Lots of code is commented out. Need to add TODO comment explaining why it was done and what should be done later

Raibo: This will be reworked within a same task, so I considered a TODO not necessary here.

PR #134, opened 2020-07-02
Added order preserving for availability search 1 step

mingazhev:


Api/Services/Accommodations/Availability/AvailabilityStorage.cs
```diff
@@ -4,26 +4,36 @@
...
 using HappyTravel.Edo.Api.Models.Availabilities;
 using HappyTravel.Edo.Common.Enums;
 using HappyTravel.EdoContracts.Accommodations;
 using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
 
 namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
 {
-    public class AvailabilityStorage : IAvailabilityStorage
+    public sealed class AvailabilityStorage : IAvailabilityStorage
```
hasmikharutyunyan, line 18: Why we need sealed here?

mingazhev: It's recommended to add this to make clear that the class will not be inherited. But in our codestyle it is redundant, removing

PR #144, opened 2020-07-09
Move test classes to similar namespaces as classes they test

Raibo:
Move test classes to similar namespaces as classes they test


HappyTravel.Edo.UnitTests/HappyTravel.Edo.UnitTests.csproj
```diff
@@ -27,8 +33,4 @@
       <ProjectReference Include="..\Api\HappyTravel.Edo.Api.csproj" />
     </ItemGroup>
 
-    <ItemGroup>
-      <Folder Include="Documents\Invoices" />
-    </ItemGroup>
```
Raibo, line 32: What is this and what to do with it?

mingazhev: This was added accidentally, I think

PR #146, opened 2020-07-13
Technical/nijo 725 entity lock with using

Raibo:


Api/Services/Payments/Accounts/AccountPaymentProcessingService.cs
```diff
@@ -389,19 +370,17 @@ private async Task<Result<PaymentAccount>> LockAccount(PaymentAccount account)
...
+        private async Task<EntityLock<TEntity>> GetEntityLock<TResult, TEntity>(Result<TResult> result, Func<Result<TResult>, TEntity> entityGetter)
+            where TEntity : IEntity =>
+            result.IsSuccess
+                ? await _locker.CreateLock<TEntity>(entityGetter(result).Id.ToString(), nameof(IAccountPaymentProcessingService))
+                : default;
 
 
-        private Task UnlockAccount(int accountId) =>
-            _locker.Release<PaymentAccount>(accountId.ToString());
+        private Result<TResult> InsureLocked<TResult, TEntity>(Result<TResult> result, EntityLock<TEntity> entityLock) =>
```
kirillta, line 380: `EnsureLocked`

kirillta, line 380: Could be made `static`

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
```diff
@@ -240,36 +223,17 @@ private async Task<Result<CounterpartyAccount>> GetCounterpartyAccount(int count
...
-                : Result.Failure<PaymentAccount>(error);
-        }
-
-
-        private async Task<Result> UnlockPaymentAccount(Result<PaymentAccount> result, int accountId)
-        {
-            await _locker.Release<PaymentAccount>(accountId.ToString());
-            return result;
-        }
+        private Result<TResult> InsureLocked<TResult, TEntity>(Result<TResult> result, EntityLock<TEntity> entityLock) =>
```
kirillta, line 233: Why didn't you move those two methods in a separate (maybe even static) class? They are pretty much the same

Raibo: There would be at least two methods for locks, not everywhere something of IEntity could be passed.
Same with EnsureLocked, two methods, one for result with value and one for result without value.

Also, we would need to pass current interface name, which will make calls bulky and kind of awkward.
We make methods for Ensure everywhere, just to avoid lambda calls in pipeline, similar approach here.

As an option, we could make static helper class, but wrap calls to it inside services and fill interface parameter name there, which will not get rid of additional methods in services, but will make them small and without major logic.

kirillta: I'm pretty sure that could be generalized without extra sufferings

kirillta: @mingazhev

Api/Services/Payments/Accounts/AccountPaymentProcessingService.cs
```diff
@@ -292,12 +361,27 @@ async Task<bool> IsRecipientAgencyChildOfPayerAgency((PaymentAccount payerAccoun
                 accounts.payerAccount.Currency == amount.Currency;
 
 
-            Task<Result<(PaymentAccount payerAccount, PaymentAccount recipientAccount)>> LockAccounts(
-                (PaymentAccount payerAccount, PaymentAccount recipientAccount) accounts) =>
-                Result.Ok()
-                    .Bind(() => LockAccount(payerAccountId))
-                    .Bind(() => LockAccount(recipientAccountId))
-                    .Map(() => accounts);
+            async Task<Result<(PaymentAccount, PaymentAccount)>> LockPayerAccount((PaymentAccount payerAccount, PaymentAccount recipientAccount) accounts)
```
hasmikharutyunyan, line 364: maybe we can separate file with this methods? As I understand everywhere this methods are the same, entity types are different

Raibo: We can't, because of the enclosure. 
We can't pass lock variable by reference because method is async, and passing it by value makes no sense.

hasmikharutyunyan: We can't in `Entitylocker` class have `Release` method and after creating lock save object into class field and than if we call `Release` method it can access to locker by class field?

Api/Infrastructure/EntityLocker.cs
```diff
@@ -18,7 +18,17 @@ public EntityLocker(EdoContext context, ILogger<EntityLocker> logger)
         }
 
 
-        public async Task<Result> Acquire<TEntity>(string entityId, string locker)
+        public Task Release<TEntity>(string entityId) => _context.RemoveEntityLock(GetEntityDescriptor<TEntity>(entityId));
+
+
+        public async Task<EntityLock<TEntity>> CreateLock<TEntity>(string entityId, string locker)
```
mingazhev, line 24: Why create, not Acquire?

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
```diff
@@ -41,21 +42,38 @@ public async Task<Result<CounterpartyBalanceInfo>> GetBalance(int counterpartyId
...
                 .Ensure(a => AreCurrenciesMatch(a, paymentData), "Account and payment currency mismatch")
                 .Ensure(IsAmountPositive, "Payment amount must be a positive number")
-                .Bind(LockCounterpartyAccount)
-                .BindWithTransaction(_context, account => Result.Ok(account)
+                .Bind(LockAccount)
+                .BindWithTransaction(_context, account => Result.Success(account)
                     .Map(AddMoneyToCounterparty)
                     .Map(WriteAuditLog))
-                .Finally(result => UnlockCounterpartyAccount(result, counterpartyAccountId));
+                .Finally(ReleaseAccount);
```
mingazhev, line 55: As it written in synonym.com, the nearest antonym for lock is "unlock"

изображение

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
```diff
@@ -112,32 +145,59 @@ async Task<CounterpartyAccount> WriteAuditLog(CounterpartyAccount account)
...
-                    .Map(WriteAuditLog))
-                .Finally(UnlockAccounts);
+                    .Tap(WriteAuditLog))
+                .Finally(ReleaseAccounts);
+
+
+            async Task<Result<CounterpartyAccount>> LockCounterpartyAccount(CounterpartyAccount account)
+            {
+                counterpartyAccountLock = await CreateEntityLock(account);
+                return Result.Success(account).Ensure(_ => counterpartyAccountLock.Acquired, counterpartyAccountLock.Error);
```
mingazhev, line 173: Why not just ternar statement there?
Result.Success(account) looks not necessary there

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
```diff
@@ -112,32 +145,59 @@ async Task<CounterpartyAccount> WriteAuditLog(CounterpartyAccount account)
...
+
             return GetCounterpartyAccount(counterpartyAccountId)
                 .Ensure(a => AreCurrenciesMatch(a, amount), "Account and payment currency mismatch")
                 .Ensure(IsAmountPositive, "Payment amount must be a positive number")
                 .Bind(LockCounterpartyAccount)
                 .Ensure(IsBalanceSufficient, "Could not charge money, insufficient balance")
                 .Bind(GetDefaultAgencyAccount)
-                .Bind(LockPaymentAccount)
-                .BindWithTransaction(_context, accounts => Result.Ok(accounts)
+                .Bind(LockAgencyAccount)
```
mingazhev, line 163: Is it really necessary to lock accounts in different stages? Why not getting locks for both at once?

Api/Infrastructure/EntityLock.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace HappyTravel.Edo.Api.Infrastructure
+{
+    public readonly struct EntityLock<TEntity> : IAsyncDisposable
+    {
+        public EntityLock(bool acquired, string entityId, string error, IEntityLocker entityLocker)
```
mingazhev, line 10: Passing "acquired" flag looks a bit strange for me. Does a an instance of "lock" have any meaning if it is not acquired?

Api/Infrastructure/EntityLock.cs
```diff
@@ -0,0 +1,31 @@
...
+            Acquired = acquired;
+            _entityId = entityId;
+            Error = error;
+            _entityLocker = entityLocker;
+        }
+
+
+        public ValueTask DisposeAsync() =>
+            Acquired
+                ? new ValueTask(_entityLocker.Release<TEntity>(_entityId))
```
mingazhev, line 21: Could we get rid of such wrapping and return ValueTask directly?

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
Raibo: Takes less space in this form

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
Raibo: Googled this a bit, looks like people out there always use specifically 'release' word when it comes to 'unlock' a lock.
Also it was already 'release' in EntityLocker class

Api/Infrastructure/EntityLocker.cs
Raibo: Because we get an instance in return, which represents the lock.
Acquire is there too, but it just acquires lock in the database.

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
Raibo: We don't need to lock agency account, if we already know that counterparty account don't hold enough money to transfer

Api/Infrastructure/EntityLock.cs
Raibo: It holds the error text.

Api/Infrastructure/EntityLock.cs
Raibo: Changed to await

Api/Services/Payments/Accounts/AccountPaymentProcessingService.cs
Raibo: If I understand it correct, we create a list of locks inside a locker instance, where we store acquired locks and then release them all when method ends.

Looks like it could save us a lot of lines of code, but it means that EntityLocker will have a state.
Such a state could make the code complex and tricky to understand.

Also, this would mean that locker could be used only by one method at the same time, which is not a great thing to rely on.

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
mingazhev: As I understand you can release a lock, not the locked entity. Anyway, there is no symmetry in lock/release

Api/Infrastructure/EntityLock.cs
mingazhev: This is not an answer for why we pass a bool flag there. Is it a lock or lock operation result?

Api/Infrastructure/EntityLocker.cs
mingazhev: Acquiring lock in the database just creates a row too))

Api/Services/Payments/Accounts/CounterpartyAccountService.cs
kirillta: I consider `release` as a valid term here

PR #153, opened 2020-07-18
BindWithLock instead of manual locking

Raibo:


Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+using HappyTravel.Edo.Data;
+
+namespace HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions
+{
+    public static class ResultLockExtensions
+    {
+        private const int FrameIndex = 8;
+
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
mingazhev, line 16: Why is this attribute there?

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using CSharpFunctionalExtensions;
+using HappyTravel.Edo.Data;
+
+namespace HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions
+{
+    public static class ResultLockExtensions
+    {
+        private const int FrameIndex = 8;
```
mingazhev, line 13: This should go after methods

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+
+        private static string GetCaller()
+        {
+            var method = new StackTrace().GetFrame(FrameIndex).GetMethod();
+            var methodDeclaringType = method.DeclaringType;
+
+            return $"{methodDeclaringType?.Name}.{method.Name}";
```
mingazhev, line 101: Did you check that method name and a class name will be readable there?

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static async Task<Result> BindWithLock<TInput>(
+            this Task<Result<TInput>> self,
+            IEntityLocker locker, Func<TInput, Task<Result>> f)
+            where TInput : IEntity
+        {
+            var (_, isFailure, entity, error) = await self.ConfigureAwait(Result.DefaultConfigureAwait);
+            return isFailure
+                ? Result.Failure(error)
+                : await WithLock(locker, GetCaller(), () => f(entity).Map(Dummy), (typeof(TInput), entity.Id.ToString()));
```
mingazhev, line 38: We have a special "VoidObject" class for such situations, please use it

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
mingazhev: And how this number was calculated?

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
mingazhev: I've checked, async methods are never inlined

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+namespace HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions
+{
+    public static class ResultLockExtensions
+    {
+        private const int FrameIndex = 8;
+
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static async Task<Result<TOutput>> BindWithLock<TInput, TOutput>(
+            this Task<Result<TInput>> self,
```
kirillta, line 18: We usually use `target` instead of `self`

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+            IEntityLocker locker, Func<TInput, Task<Result<TOutput>>> f)
```
kirillta, line 19: Also avoid one letter names for input parameters. I suggest `func` here

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
Raibo: By checking whats on every stack layer :)

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
Raibo: Yes, when ran tests.
The tests are further in stack than actual services, so it should be OK

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
Raibo: I took `ResultTransactionExtensions` as a reference.
Should I still rename this?

Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
Raibo: Same here, `ResultTransactionExtensions` was my reference. 
Do we need this renamed in both classes?

kirillta: Highly desirable for both cases

PR #157, opened 2020-07-19
Added disallowing all robots.txt for the api

mingazhev:
Due to sentry error:
https://sentry.dev.happytravel.com/happytravel/edo-api/issues/987/


kirillta:
Ho-ho

PR #158, opened 2020-07-19
Replaced memory flow with double and distributed flow where possible

mingazhev:
+ Updated FloxDc.* versions

### DIFF
--- a/Api/Services/CurrencyConversion/CurrencyRateService.cs
+++ b/Api/Services/CurrencyConversion/CurrencyRateService.cs
@@ -17,20 +17,20 @@ namespace HappyTravel.Edo.Api.Services.CurrencyConversion
     {
         public CurrencyRateService(IHttpClientFactory httpClientFactory,
             IOptions<CurrencyRateServiceOptions> options,
-            IMemoryFlow memoryFlow)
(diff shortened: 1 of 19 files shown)


Api/Services/CurrencyConversion/CurrencyRateService.cs
```diff
@@ -17,20 +17,20 @@ public class CurrencyRateService : ICurrencyRateService
...
-            IMemoryFlow memoryFlow)
+            IDoubleFlow memoryFlow)
         {
             _httpClientFactory = httpClientFactory;
             _memoryFlow = memoryFlow;
             _options = options.Value;
         }
 
-        public ValueTask<Result<decimal>> Get(Currencies source, Currencies target)
+        public async ValueTask<Result<decimal>> Get(Currencies source, Currencies target)
```
kirillta, line 27: Is this `async` necessary?

mingazhev: Yes, because doubleflow returns a Task<T>, not ValueTask. So to leave an ability for fast return (without allocation) this task is awaited, not returned

PR #159, opened 2020-07-19
Moved getting agent context to controllers for bookings and removed s…

mingazhev:
…etting agent context in async response processors

There was a problem:
Because of implicit getting AgentContext deep inside in services some outer services (like async booking response processors) could fail, because request to them does not contain any information about the agent context.

I've moved getting agent context up to controllers for booking services and removed setting this implicit context.


kirillta:
Reasonable solution. I wonder if we'll have tests on those classes someday

PR #162, opened 2020-07-19
Feature/nijo 331 search speed improvements

mingazhev:
1. Saving state to storage and starting a search in parallel
2. Added caching for agent permissions
3. Little cleanup: Async methods in location service, currency rate service naming
4. Caching agent settings for markup application

### DIFF
--- a/Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
+++ b/Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
@@ -123,8 +123,11 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
             async Task<Result<AvailabilityDetails, ProblemDetails>> GetAvailability(EdoContracts.Accommodations.AvailabilityRequest request,
                 string languageCode)
             {
-                await storage.SetState(searchId, providerKey, AvailabilitySearchState.Pending(searchId));
-                return await dataProvider.GetAvailability(request, languageCode);
+                var saveToStorageTask = storage.SetState(searchId, providerKey, AvailabilitySearchState.Pending(searchId));
+                var getAvailabilityTask = dataProvider.GetAvailability(request, languageCode);
+                await Task.WhenAll(saveToStorageTask, getAvailabilityTask);
+
+                return getAvailabilityTask.GetAwaiter().GetResult();
             }
 
 
--- a/Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
+++ b/Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
@@ -127,7 +127,7 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
                 var getAvailabilityTask = dataProvider.GetAvailability(request, languageCode);
                 await Task.WhenAll(saveToStorageTask, getAvailabilityTask);
 
-                return getAvailabilityTask.GetAwaiter().GetResult();
+                return getAvailabilityTask.Result;
             }
 
 
--- a/Api/Services/Agents/PermissionChecker.cs
+++ b/Api/Services/Agents/PermissionChecker.cs
@@ -23,7 +23,7 @@ namespace HappyTravel.Edo.Api.Services.Agents
 
         public async ValueTask<Result> CheckInAgencyPermission(AgentContext agent, InAgencyPermissions permission)
         {
-            var key = _flow.BuildKey(nameof(PermissionChecker), "Permissions", 
+            var key = _flow.BuildKey(nameof(PermissionChecker), nameof(CheckInAgencyPermission), 
                 agent.AgencyId.ToString(),
                 agent.AgentId.ToString());
             
--- a/Api/Services/Agents/PermissionChecker.cs
+++ b/Api/Services/Agents/PermissionChecker.cs
@@ -1,6 +1,9 @@
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
+using FloxDc.CacheFlow;
+using FloxDc.CacheFlow.Extensions;
 using HappyTravel.Edo.Api.Models.Agents;
 using HappyTravel.Edo.Common.Enums;
 using HappyTravel.Edo.Data;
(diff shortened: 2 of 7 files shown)


Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
```diff
@@ -123,8 +123,11 @@ await GetAvailability(request, languageCode)
             async Task<Result<AvailabilityDetails, ProblemDetails>> GetAvailability(EdoContracts.Accommodations.AvailabilityRequest request,
                 string languageCode)
             {
-                await storage.SetState(searchId, providerKey, AvailabilitySearchState.Pending(searchId));
-                return await dataProvider.GetAvailability(request, languageCode);
+                var saveToStorageTask = storage.SetState(searchId, providerKey, AvailabilitySearchState.Pending(searchId));
+                var getAvailabilityTask = dataProvider.GetAvailability(request, languageCode);
+                await Task.WhenAll(saveToStorageTask, getAvailabilityTask);
+
+                return getAvailabilityTask.GetAwaiter().GetResult();
```
kirillta, line 130: The `Result` is enough here, because a task is completed with a guarantee

kirillta: It's a matter of discuss if we'll have improvements over parallel multi-user queries

Api/Services/Agents/PermissionChecker.cs
```diff
@@ -10,28 +13,47 @@ namespace HappyTravel.Edo.Api.Services.Agents
...
+            var storedPermissions = await _flow.GetOrSetAsync(
+                key: key, 
+                getValueFunction: async () => await GetPermissions(agent.AgentId, agent.AgencyId), 
+                AgentPermissionsCacheLifeTime);
 
             if (Equals(storedPermissions, default))
                 return Result.Failure("The agent isn't affiliated with the agency");
 
             return !storedPermissions.HasFlag(permission)
                 ? Result.Failure($"Agent does not have permission '{permission}'")
```
kirillta, line 39: I suggest $"Agent does not have the '{permission}' permission"

Api/Services/Agents/PermissionChecker.cs
```diff
@@ -10,28 +13,47 @@ namespace HappyTravel.Edo.Api.Services.Agents
...
 
 
         public async ValueTask<Result> CheckInAgencyPermission(AgentContext agent, InAgencyPermissions permission)
         {
-            var storedPermissions = await _context.AgentAgencyRelations
-                .Where(r => r.AgentId == agent.AgentId)
-                .Where(r => r.AgencyId == agent.AgencyId)
-                .Select(r => r.InAgencyPermissions)
-                .SingleOrDefaultAsync();
+            var key = _flow.BuildKey(nameof(PermissionChecker), "Permissions",
```
kirillta, line 26: I suggest to use `nameof(CheckInAgencyPermission)` instead of `"Permissions"`, it's less vulnerable to misusing

Api/Services/Markups/MarkupService.cs
```diff
@@ -149,6 +163,7 @@ string BuildKey(MarkupPolicy policyWithFunc)
 
         private static readonly TimeSpan MarkupPolicyFunctionCachingTime = TimeSpan.FromDays(1);
         private static readonly TimeSpan AgentPoliciesCachingTime = TimeSpan.FromMinutes(5);
+        private static readonly TimeSpan AgentSettingsCachingTime = TimeSpan.FromMinutes(5);
```
kirillta, line 166: Why the caching time is different from the other places where you set 2 minutes?

Api/Services/CurrencyConversion/CurrencyRateService.cs
```diff
@@ -17,10 +17,10 @@ public class CurrencyRateService : ICurrencyRateService
     {
```
kirillta, line 17: Doubt we're winning anything here, like the value cached on the other side, and we replacing one network request to the converter with another network request to a cache instance

Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
mingazhev: Yes, this can increase the request speed by decreasing a throughput a bit. But we need better request speed now, don't we?

Api/Services/CurrencyConversion/CurrencyRateService.cs
mingazhev: I thought that we can get value from a memory in this case, without making any network requests. Or is network request is executed everytime?

Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
mingazhev: Changed the task to just returning result

Api/Services/Agents/PermissionChecker.cs
mingazhev: Fixed

Api/Services/Agents/PermissionChecker.cs
mingazhev: Fixed

Api/Services/Markups/MarkupService.cs
mingazhev: Set to 2 minutes

Api/Services/CurrencyConversion/CurrencyRateService.cs
kirillta: Then `IMemoryFlow` is enough

Api/Services/Markups/MarkupService.cs
```diff
@@ -149,6 +163,7 @@ string BuildKey(MarkupPolicy policyWithFunc)
 
         private static readonly TimeSpan MarkupPolicyFunctionCachingTime = TimeSpan.FromDays(1);
         private static readonly TimeSpan AgentPoliciesCachingTime = TimeSpan.FromMinutes(5);
+        private static readonly TimeSpan AgentSettingsCachingTime = TimeSpan.FromMinutes(2);
```
kirillta, line 166: Now there is a difference between `AgentPoliciesCachingTime` and `AgentSettingsCachingTime` caching intervals

Api/Services/Markups/MarkupService.cs
```diff
@@ -48,6 +48,20 @@ public async Task<Markup> Get(AgentContext agentContext, MarkupPolicyTarget poli
...
+        private Task<AgentUserSettings> GetAgentSettings(AgentContext agentContext)
+        {
+            return _flow.GetOrSetAsync(
+                key: BuildKey(),
+                getValueFunction: async () => await _agentSettingsManager.GetUserSettings(agentContext),
+                AgentSettingsCachingTime); 
+            
+            string BuildKey()
+                => _flow.BuildKey(nameof(MarkupService),
+                    "AgentSettings",
```
kirillta, line 60: A hard-coded value again

Api/Services/Accommodations/Availability/AvailabilitySearchScheduler.cs
kirillta: > Yes, this can increase the request speed by decreasing a throughput a bit. But we need better request speed now, don't we?

We do

Api/Services/Markups/MarkupService.cs
mingazhev: Removed from everywhere where found

Api/Services/Markups/MarkupService.cs
mingazhev: Set them same

Api/Services/CurrencyConversion/CurrencyRateService.cs
mingazhev: Let's leave it as is for now

kirillta: Ok, but create a task for that

PR #163, opened 2020-07-20
Feature/ueda 6 autocomplete for counterparties

hasmikharutyunyan:


Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -66,6 +67,25 @@ select new
...
+        public Task<List<CounterpartyPrediction>> GetCounterpartiesPredictions(string query)
+            => (from c in _context.Counterparties
+                    join ag in _context.Agencies on c.Id equals ag.CounterpartyId
+                    join ar in _context.AgentAgencyRelations on ag.Id equals ar.AgencyId
+                    join a in _context.Agents on ar.AgentId equals a.Id
+                    where c.IsActive
+                        && a.IsActive
+                        && ar.Type == AgentAgencyRelationTypes.Master
+                        && c.State == CounterpartyStates.FullAccess
+                        && (EF.Functions.ILike(c.Name, $"%{query}%")
```
kirillta, line 79: Those could be `$"{query}%"`

Api/Models/Management/Enums/AdministratorPermissions.cs
```diff
@@ -15,5 +15,6 @@ public enum AdministratorPermissions
         CounterpartyBalanceReplenishAndSubtract = 128,
         CounterpartyToAgencyTransfer = 256,
         CounterpartyManagement = 512,
+        PaymentLinkGeneration =1024
```
kirillta, line 18: Must check if a new counterparty master gets this permission on verification

hasmikharutyunyan: This is for administrator

PR #172, opened 2020-07-30
Bugfix/nijo 778 return valid permissions

mingazhev:
First I've started from place where we should return actual permissions to user, but backend returns all.
Than I've moved some methods and made passing agent context explicitly in some service methods


kirillta:
Do we still have a method which returns a full list of existing permissions?

mingazhev:
> 
> 
> Do we still have a method which returns a full list of existing permissions?

Yes we do, this is needed on frontend side

PR #179, opened 2020-08-09
Added parameter to ignore provider's error during booking cancellatio…

Raibo:


Api/Controllers/AgentControllers/AccommodationsController.cs
```diff
@@ -270,6 +275,27 @@ public async Task<IActionResult> CancelBooking(int bookingId)
         }
 
 
+        /// <summary>
```
mingazhev, line 278: This method should be placed inside the admin controllers folder, maybe create a separate controller for it

Api/Controllers/AgentControllers/AccommodationsController.cs
```diff
@@ -270,6 +275,27 @@ public async Task<IActionResult> CancelBooking(int bookingId)
...
+        ///     Cancel accommodation booking by admin.
+        /// </summary>
+        /// <param name="bookingId">Id of booking to cancel</param>
+        /// <param name="ignoreProviderError">If a provider returns an error after cancellation request, this is ignored as if it was a success</param>
+        /// <returns></returns>
+        [HttpPost("accommodations/bookings/{bookingId}/cancel-by-admin")]
```
mingazhev, line 284: After moving just "cancel" verb will be enough

Api/Services/Accommodations/Bookings/BookingService.cs
```diff
@@ -296,8 +296,18 @@ async Task NotifyAgent()
 
             return await ProcessBookingCancellation(booking, serviceAccount.ToUserInfo());
         }
-        
-        
+
+
+        public async Task<Result<VoidObject, ProblemDetails>> Cancel(int bookingId, Administrator administrator, bool ignoreProviderError)
```
kirillta, line 301: The name `ignoreProviderError` is unclear. May be `requireExistingBooking` or `doesBookingExist`. Also not clear when I must use the condition

Api/Models/Management/Enums/AdministratorPermissions.cs
```diff
@@ -17,5 +17,6 @@ public enum AdministratorPermissions
         CounterpartyManagement = 512,
         PaymentLinkGeneration = 1024,
         AccommodationDuplicatesReportApproval = 2048,
+        BoookingCancellation = 4096,
```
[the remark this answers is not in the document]
Raibo: I thought we don't store any admin permissions yet

Api/Services/Accommodations/Bookings/BookingService.cs
Raibo: This flag has nothing to do with booking exists or not, it tells whether pipeline should fail on connector error after cancellation attempt or not.
Can't really think of any better name here.

Api/Models/Management/Enums/AdministratorPermissions.cs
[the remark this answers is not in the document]
kirillta: Ok

Api/Services/Accommodations/Bookings/BookingService.cs
kirillta: At least it must explain why a booking requires and why I want skip it

PR #184, opened 2020-08-13
4/5 Feature/nijo 787 reimplement second step

mingazhev:
Changed the second step to support discussed logic.
For now it does simply unites results from different providers


Api/Controllers/AgentControllers/AccommodationsController.cs
```diff
@@ -47,35 +47,12 @@ public class AccommodationsController : BaseController
...
-        }
-
-
         /// <summary>
         ///     Starts availability search and returns an identifier to fetch results later
         /// </summary>
         /// <param name="request">Availability request</param>
         /// <returns>Search id</returns>
-        [HttpPost("availabilities/accommodations/searches")]
+        [HttpPost("accommodations/availabilities/searches")]
```
kirillta, line 55: That was intentional

mingazhev: So we should discuss the final endpoint addresses.
Now we have `accommodations/bookings` but `availabilities/accommodations`

kirillta: `accommodation` is a root entity

PR #185, opened 2020-08-13
5/5 Changed third step availability search to support ids from previous s…

mingazhev:
Reimplemented it to use results from second step


Api/Controllers/AgentControllers/AccommodationsController.cs
```diff
@@ -29,15 +30,13 @@ namespace HappyTravel.Edo.Api.Controllers.AgentControllers
     [Produces("application/json")]
     public class AccommodationsController : BaseController
     {
-        public AccommodationsController(IAccommodationService service, 
-            IFirstStepAvailabilitySearchService firstStepAvailabilitySearchService,
+        public AccommodationsController(IFirstStepAvailabilitySearchService firstStepAvailabilitySearchService,
```
kirillta, line 33: Still not renamed

mingazhev: Fixed

PR #188, opened 2020-08-20
Bugfix/nijo 817 validate counterparty approval states

mingazhev:
As discussed with @kirillta , valid state changes are:
**PendingVerification->VerifiedReadOnly
VerifiedReadOnly->VerifiedFullAccess
PendingVerification->FailedVerification**

All others should be prohibited.
Additionally, removed one hack, added setting verification in the same transaction where accounts are created

### DIFF
--- a/Api/AdministratorServices/CounterpartyManagementService.cs
+++ b/Api/AdministratorServices/CounterpartyManagementService.cs
@@ -145,27 +145,34 @@ namespace HappyTravel.Edo.Api.AdministratorServices
         }
 
 
-        public Task<Result> VerifyAsFullyAccessed(int counterpartyId, string verificationReason)
+        public async Task<Result> VerifyAsFullyAccessed(int counterpartyId, string verificationReason)
         {
-            return Verify(counterpartyId, counterparty => Result.Ok(counterparty)
+            return await GetCounterparty(counterpartyId)
                 .Ensure(c => c.State == CounterpartyStates.ReadOnly,
-                    "Verification as fully accessed is only available for companies that were verified as read-only earlier")
-                .Tap(c => SetVerified(c, CounterpartyStates.FullAccess, verificationReason))
-                .Bind(_ => Task.FromResult(Result.Ok())) // HACK: conversion hack because can't map tasks
-                .Tap(() => WriteToAuditLog(counterpartyId, verificationReason)));
+                    "Verification as fully accessed is only available for counterparties that were verified as read-only earlier")
+                .Tap(c => SetVerificationState(c, CounterpartyStates.FullAccess, verificationReason))
+                .Tap(() => WriteToAuditLog(counterpartyId, verificationReason));
         }
 
 
         public Task<Result> VerifyAsReadOnly(int counterpartyId, string verificationReason)
         {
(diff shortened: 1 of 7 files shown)


Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -186,6 +193,16 @@ async Task<Result> CreateAccountsForAgencies(Counterparty counterparty)
             }
         }
 
+        
+        public async Task<Result> FailVerification(int counterpartyId, string verificationReason)
+        {
+            return await GetCounterparty(counterpartyId)
+                .Ensure(c => c.State == CounterpartyStates.PendingVerification,
+                    "Verification failure is only available for counterparties that are in a pending state")
+                .Tap(c => SetVerificationState(c, CounterpartyStates.FailedVerification, verificationReason))
```
hasmikharutyunyan, line 202: Maybe `UnVerified` is better ? For me `FailedVerification` state is asociated with that,when we wanted verify but it is not happen.

mingazhev: Unverified looks for me same as PendingVerification. In meaning "not verified yet" but maybe will be verified later

Raibo: How about RejectedVerification or DeclinedVerification?

hasmikharutyunyan: I don't think so, but maybe `NoAccess` ?

Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -145,27 +145,34 @@ private async Task<Result<Counterparty>> GetCounterparty(int counterpartyId)
...
-            return Verify(counterpartyId, counterparty => Result.Ok(counterparty)
+            return await GetCounterparty(counterpartyId)
                 .Ensure(c => c.State == CounterpartyStates.ReadOnly,
-                    "Verification as fully accessed is only available for companies that were verified as read-only earlier")
-                .Tap(c => SetVerified(c, CounterpartyStates.FullAccess, verificationReason))
-                .Bind(_ => Task.FromResult(Result.Ok())) // HACK: conversion hack because can't map tasks
-                .Tap(() => WriteToAuditLog(counterpartyId, verificationReason)));
+                    "Verification as fully accessed is only available for counterparties that were verified as read-only earlier")
+                .Tap(c => SetVerificationState(c, CounterpartyStates.FullAccess, verificationReason))
+                .Tap(() => WriteToAuditLog(counterpartyId, verificationReason));
```
Raibo, line 154: Do we need `BindWithTransaction` here like with payments to ensure successful audit log operation?

Api/AdministratorServices/CounterpartyManagementService.cs
mingazhev: @kirillta what do you think?
DeclinedVerification
NoAccess 

both good for me

Api/AdministratorServices/CounterpartyManagementService.cs
mingazhev: Not required much but maybe useful because there are a lot of operations that are good to undo at once - such as accounts creation

PR #189, opened 2020-08-20
Showing (not actual) saved permissions in UI

mingazhev:


Api/Services/Agents/AgentService.cs
```diff
@@ -145,7 +145,7 @@ public async Task<Result<AgentInfoInAgency>> GetAgent(int agencyId, int agentId,
                         on agency.CounterpartyId equals counterparty.Id
                     where cr.AgencyId == agencyId && cr.AgentId == agentId
                     select (AgentInfoInAgency?) new AgentInfoInAgency(agent.Id, agent.FirstName, agent.LastName, agent.Email, agent.Title, agent.Position, counterparty.Id, counterparty.Name,
-                        cr.AgencyId, agency.Name, cr.Type == AgentAgencyRelationTypes.Master, GetActualPermissions(counterparty.State, cr.InAgencyPermissions)))
+                        cr.AgencyId, agency.Name, cr.Type == AgentAgencyRelationTypes.Master, cr.InAgencyPermissions.ToList()))
```
Raibo, line 148: Is `GetActualPermissions` completely unused now?

kirillta (approved):

Api/Services/Agents/AgentService.cs
mingazhev: No, it's used when we get info about current agent.
And this data is used on frontend side where frontend can use it in UI logic

PR #193, opened 2020-08-24
Fix caller method and service search in ResultLockExtensions

Raibo:
Before changes, caller method from a service was searched in a set stack frame.
Now a cycle checks stack frames one by one and compares caller class to have "Service" in the end of its name.


Api/Infrastructure/FunctionalExtensions/ResultLockExtensions.cs
```diff
@@ -86,19 +86,32 @@ public static class ResultLockExtensions
...
 
+                if (declaringType.Name.EndsWith("Service"))
+                    return $"{declaringType.Name}.{method.Name}";
+
+                i++;
+            }
 
-        private const int FrameIndex = 8;
+            return "UnknownCallerService";
+        }
```
Raibo, line 112: The cycle should not hang because stackTrace.GetFrame(i) when i > number of frames returns null.

kirillta (approved):

PR #194, opened 2020-08-28
Improved logging during operations with bookings

Raibo:


Raibo:
Comment for @oleg1std : in case of failure during refund to account in AccountPaymentProcessingService.RefundMoney, logs will be written, but not in AccountPaymentProcessingService directly, instead in BookingService. 
This refund method is only called from BookingService.

kirillta (approved):

PR #198, opened 2020-09-01
changed payment history routes

hasmikharutyunyan:
There was bug that we can not see payment history, tha reason was, that from frontend instead of agencyId we got counterpartyId. 
Bug would not be if routs were more understandable.


kirillta (approved):

mingazhev (approved):
Need to make corresponding changes on frontend side

PR #200, opened 2020-09-02
Enabled connectors for agent are got from database

mingazhev:
This is the second part.
We do have separate setting for each agent for enabled connectors.
In this PR there is no API to set it by admin, this will be added soon or maybe much later.
For now we can adjust it for only one user (Denis) in database by hand.

What this is needed for?
We urgently need to enable Illusions connector on prod, just for one user, for testing.

### DIFF
--- a/HappyTravel.Edo.Data/Agents/AvailabilitySearchSettings.cs
+++ b/HappyTravel.Edo.Data/Agents/AvailabilitySearchSettings.cs
@@ -9,7 +9,7 @@ namespace HappyTravel.Edo.Data.Agents
         [JsonConstructor]
         public AvailabilitySearchSettings(List<DataProviders> enabledProviders)
         {
-            EnabledProviders = enabledProviders;
+            EnabledProviders = enabledProviders ?? new List<DataProviders>();
         }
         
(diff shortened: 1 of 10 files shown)


HappyTravel.Edo.Data/Agents/AvailabilitySearchSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Data.Agents
+{
+    public readonly struct AvailabilitySearchSettings
+    {
+        [JsonConstructor]
+        public AvailabilitySearchSettings(List<DataProviders> enabledProviders)
+        {
+            EnabledProviders = enabledProviders;
```
kirillta, line 12: Here must be either a default, either a null exception

kirillta (approved):

HappyTravel.Edo.Data/Agents/AvailabilitySearchSettings.cs
mingazhev: Fixed

PR #202, opened 2020-09-03
Settings agent enabled connectors by an admin

mingazhev:


Api/Controllers/AdministratorControllers/AgentsController.cs
```diff
@@ -0,0 +1,43 @@
...
+    [Produces("application/json")]
+    public class AgentsController : BaseController
+    {
+        public AgentsController(IAgentSystemSettingsManagementService systemSettingsManagementService)
+        {
+            _systemSettingsManagementService = systemSettingsManagementService;
+        }
+        
+        
+        [HttpPut("agencies/{agencyId}/agents/{agentId}/system-settings/availability-search")]
```
kirillta, line 27: No method descriptions

kirillta (approved):

Api/Controllers/AdministratorControllers/AgentsController.cs
mingazhev: Fixed

PR #203, opened 2020-09-03
moved get deadline request to the separate service and added check ro…

hasmikharutyunyan:
…omContractSet on second step cache

### DIFF
--- a/Api/Services/Accommodations/Availability/DeadlineService.cs
+++ b/Api/Services/Accommodations/Availability/DeadlineService.cs
@@ -28,28 +28,46 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
             string languageCode)
         {
             var enabledProviders = await _dataProviderManager.GetEnabled(agent);
-            
-            var selectedResult = (await _availabilityStorage.GetResults(searchId, enabledProviders))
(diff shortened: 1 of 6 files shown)


Api/Services/Accommodations/Availability/DeadlineService.cs
```diff
@@ -0,0 +1,60 @@
...
+            _dataProviderManager = dataProviderManager;
+        }
+
+
+        public async Task<Result<Deadline, ProblemDetails>> GetDeadlineDetails(Guid searchId, Guid resultId, Guid roomContractSetId, AgentContext agent,
+            string languageCode)
+        {
+            var enabledProviders = await _dataProviderManager.GetEnabled(agent);
+            
+            var selectedResult = (await _availabilityStorage.GetResults(searchId, enabledProviders))
```
mingazhev, line 32: Could you reorganize this method to separate ones to make the main logic clear?
I see that we:
1. Search for result in wideAvailabilityStorage. Return if found
2. If result is not found search, search in roomSelectionStorage. Return if found.

In my opinion, these are good candidates for separate methods - GetFromAvailabilityStorage and GetFromRoomSelectionStorage (or something similar)

mingazhev (approved):

PR #215, opened 2020-09-17
Show not paid bookings with account payment method

Raibo:


Api/Services/Accommodations/Bookings/BookingRecordsManager.cs
```diff
@@ -200,7 +201,8 @@ public async Task<Result<List<SlimAccommodationBookingInfo>>> GetAgentBookingsIn
         {
             var bookingData = await _context.Bookings
                 .Where(b => b.AgentId == agentContext.AgentId)
-                .Where(b => b.PaymentStatus != BookingPaymentStatuses.NotPaid)
+                .Where(b => b.PaymentMethod == PaymentMethods.BankTransfer
```
mingazhev, line 209: Should there be the brackets?

Raibo: No, it's good like this.
The 'and' (&&) operation occurs earlier than 'or' (||) operation, like '*' and '+'. 
A statement like this `var testBool = true || false && false;` ends up true.

mingazhev: Anyway, this is hard to read and understand the logic at one glance

Raibo: Added brackets and newline.

mingazhev (approved):

PR #219, opened 2020-09-18
Feature/nijo 747 agency payment method endpoint

Raibo:


Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -51,7 +53,50 @@ public async Task<Result<AgencyAvailabilitySearchSettings>> GetAvailabilitySearc
...
+                .Tap(SetOptions);
+
+
+            async Task SetOptions()
+            {
+                var systemSettings = await _context.AgencySystemSettings.SingleOrDefaultAsync(s => s.AgencyId == agencyId)
+                    ?? new AgencySystemSettings { AgencyId = agencyId };
+
+                systemSettings.DisplayedPaymentOptions = settings;
+                _context.Update(systemSettings);
```
mingazhev, line 71: Did you check this?
Looks like not working

mingazhev (changes requested):

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
Raibo: How could I possibly check this, if the migration will be only executed upon merging :(

mingazhev: You can execute migration, check it and then revert

mingazhev: look for EF cli dotnet tool

Raibo: I looked closely to this bit of code and immediately it became obvious why this won't work.
Fixed. 
I did not apply the migration, but given the logic in `SetAvailabilitySearchSettings` works, all should be fine.

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -51,7 +53,50 @@ public async Task<Result<AgencyAvailabilitySearchSettings>> GetAvailabilitySearc
...
+                var options = systemSettings?.DisplayedPaymentOptions;
+
+                return options == null
+                    ? Result.Failure<DisplayedPaymentOptionsSettings>("No value found for DisplayedPaymentOptions settings")
+                    : Result.Success(options.Value);
+            }
+        }
+
+
+        private Task<bool> IsAgencyExist(int agencyId) => _context.Agencies.AnyAsync(a => a.Id == agencyId && a.IsActive);
```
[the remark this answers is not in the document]
Raibo: Did as you suggest. 
Now `GetAvailabilitySearchSettings` and `SetAvailabilitySearchSettings` checks existance of an agency among active ones. Is this OK?

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -51,7 +53,50 @@ public async Task<Result<AgencyAvailabilitySearchSettings>> GetAvailabilitySearc
                 ? Result.Failure<AgencyAvailabilitySearchSettings>($"Could not find availability search settings for agency with id {agencyId}")
                 : existingSettings.AvailabilitySearchSettings;
         }
-        
+
+
+        public async Task<Result> SetDisplayedPaymentOptions(DisplayedPaymentOptionsSettings settings, int agencyId)
```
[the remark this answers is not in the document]
Raibo: This sounds not like "a deeper approach", but as a different one.
I think this should be discussed.

mingazhev (approved):

PR #226, opened 2020-09-22
Ability to pay with credit card for BankTransfer bookings

Raibo:
.

### DIFF
--- a/Api/Services/Accommodations/Bookings/BookingPaymentService.cs
+++ b/Api/Services/Accommodations/Bookings/BookingPaymentService.cs
@@ -248,27 +248,6 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Bookings
         }
 
 
-        public async Task<Result> SetPaymentMethodToCreditCard(string referenceCode)
-        {
-            return await _recordsManager.Get(referenceCode)
-                .Tap(SetPaymentMethod);
-
-
-            async Task SetPaymentMethod(Booking booking)
-            {
-                if (booking.PaymentMethod == PaymentMethods.CreditCard)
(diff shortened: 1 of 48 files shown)


Api/Services/Payments/IPaymentsService.cs
```diff
@@ -12,5 +12,7 @@ public interface IPaymentsService
         Task<Result> ProcessPaymentChanges(Payment payment);
         
         Task<Result<(int AgentId, int AgencyId)>> GetServiceBuyer(string referenceCode);
+
+        Task<Result> SetPaymentMethodToCreditCard(string referenceCode);
```
mingazhev, line 16: Should this be a separate operation?
I thought that this should be processed during credit card payment

Raibo: This is a separate operation, which is called during credit card operation

Api/Services/Accommodations/Bookings/BookingPaymentService.cs
```diff
@@ -248,6 +248,27 @@ public async Task<Result<(int AgentId, int AgencyId)>> GetServiceBuyer(string re
         }
 
 
+        public async Task<Result> SetPaymentMethodToCreditCard(string referenceCode)
```
mingazhev, line 251: I think this method will be more suitable for BookingRecordsManager and should look like "SetPaymentMethod(PaymentMethods paymentMethod)".
It just updates a booking field, nothing about payment processing

kirillta (approved):
Approved with @mingazhev comments in mind

mingazhev (approved):

PR #257, opened 2020-10-14
View all agency bookings

mingazhev:


Api/Models/Agents/AgentRelatedData.cs
```diff
@@ -0,0 +1,24 @@
...
+        public AgentRelatedData(SlimAgentDescription agent, TData data)
+        {
+            Agent = agent;
+            Data = data;
+        }
+        
+        /// <summary>
+        /// Nested data
+        /// </summary>
+        public TData Data { get; }
```
kirillta, line 17: This part isn't clear. What data?

mingazhev: There could be any data. In this request it is booking data, then it could be payment data and etc

HappyTravel.Edo.Data/Migrations/20201014172454_UpdateAgentsAllPermissions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace HappyTravel.Edo.Data.Migrations
+{
+    public partial class UpdateAgentsAllPermissions : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("UPDATE \"AgentAgencyRelations\" SET \"InAgencyPermissions\" = 4094 WHERE \"InAgencyPermissions\" =  2046");
+
```
oleg1std, line 10: Extra line

mingazhev: Fixed

Api/Models/Agents/AgentRelatedData.cs
mingazhev: Changed to AgentBoundedData

PR #275, opened 2020-10-30
Refund credit card payments with payfort

Raibo:


Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
```diff
@@ -83,6 +84,40 @@ Task WriteAuditLog(CreditCardVoidResult voidResult)
             }
         }
 
+
+        public async Task<Result<CreditCardRefundResult>> Refund(CreditCardRefundMoneyRequest request,
+            CreditCardPaymentInfo paymentInfo,
+            string maskedNumber,
+            string referenceCode,
+            UserInfo user,
+            int agentId)
```
mingazhev, line 93: Why refund is in "CaptureService"? There are two anthonym operations there - to operate with authorized money. Refund looks like not suitable there

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
```diff
@@ -83,6 +84,40 @@ Task WriteAuditLog(CreditCardVoidResult voidResult)
...
+        {
+            return await RefundInPayfort()
+                .Tap(WriteAuditLog);
+
+            async Task<Result<CreditCardRefundResult>> RefundInPayfort() => 
+                request.Amount.IsGreaterThan(0m)
+                    ? await _payfortService.Refund(request)
+                    : Result.Success(default(CreditCardRefundResult));
```
mingazhev, line 101: Result could be omitted there

mingazhev: And returning default value looks not valid there.
If we did not do anything in payfort, anyway we need to return a valid object, containing all the refund data, maybe with zero amount, not default value.
In case of default, at least useless data is written to audit log

Api/Services/Payments/Payfort/PayfortService.cs
```diff
@@ -168,6 +168,74 @@ Result<CreditCardVoidResult> CreateResult(PayfortVoidResponse model)
         }
 
 
+        public async Task<Result<CreditCardRefundResult>> Refund(CreditCardRefundMoneyRequest moneyRequest)
+        {
+            try
+            {
+                var requestContent = GetSignedContent();
+                var client = _clientFactory.CreateClient(HttpClientNames.Payfort);
+                using (var response = await client.PostAsync(_options.PaymentUrl, requestContent))
```
mingazhev, line 177: Please use the modern "using" syntax for new code

mingazhev (changes requested):

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
Raibo: Well, refund is basically a cancel of capture, so I thought it would suit here, but if you insist, I'll make new service for this operation

kirillta: I agree, but do we need another name for the service?

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
Raibo: There is some additional detail to this part of code, which matters:
`CreditCardRefundResult` object fields make only a part of the audit record. Money amount is taken not from it, but from `CreditCardRefundMoneyRequest` object.
`CreditCardRefundResult` is responsible only for payfort response, which never happened in case of zero amount, so it is obvious that no values could be substituted there. 

The audit record is not useless, it contains additional data, as mentioned above.
Also, added `MerchantReference` to returned `CreditCardRefundResult` for more complete record, since it is known even without payfort response.

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
Raibo: How about `CreditCardCaptureVoidRefundService`?

mingazhev: Ok for another service name, ok for separate service.
Not Ok for "CreditCardCaptureVoidRefundService" and other names with enumerating method names.

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
mingazhev: You're right, it's all good there

Api/Services/Payments/CreditCards/CreditCardMoneyCaptureService.cs
Raibo: Couldn't come up with any good name, so moved to "RefundService"

mingazhev (approved):

kirillta (approved):

PR #281, opened 2020-11-06
Feature/nijo 1023 added agency and counterparty activation

hasmikharutyunyan:


Api/Controllers/AdministratorControllers/CounterpartiesController.cs
```diff
@@ -139,17 +140,18 @@ public async Task<IActionResult> DeactivateCounterparty(int counterpartyId)
...
 
         /// <summary>
-        ///  Deactivates specified agency.
+        ///  Changes specified agency activity status.
         /// </summary>
         /// <param name="agencyId">Id of the agency.</param>
+        /// <param name="request">Request data for activity status change.</param>
         /// <returns></returns>
-        [HttpPut("agencies/{agencyId}/deactivate")]
+        [HttpPut("agencies/{agencyId}/change-activity-status")]
```
kirillta, line 148: Don't put an action and its subject in one route section. Should be like `activity-status/change`. But also I don't like word _activity_ here

Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -212,27 +213,31 @@ async Task<Result> CreateAccountsForAgencies(Counterparty counterparty)
...
-        public Task<Result> DeactivateCounterparty(int counterpartyId)
+
+        public Task<Result> ChangeCounterpartyActivityStatus(int counterpartyId, ActivityStatus status, string reason)
             => GetCounterparty(counterpartyId)
-                .Ensure(counterparty => counterparty.IsActive, "Counterparty already deactivated.")
-                .BindWithTransaction(_context, Deactivate);
+                .Ensure(_ => !string.IsNullOrWhiteSpace(reason), "Reason must not be empty")
+                .Ensure(counterparty => counterparty.IsActive != ConvertToDbStatus(status), $"Counterparty status is already {status}.")
+                .BindWithTransaction(_context, counterparty => ChangeActivityStatus(counterparty, status))
+                .Tap(() => WriteCounterpartyActivityStatusChangeToAuditLog(counterpartyId, reason));
```
kirillta, line 232: Shouldn't that line be within transaction?

kirillta (changes requested):

Api/AdministratorServices/CounterpartyManagementService.cs
hasmikharutyunyan: I think write log can be independent operation

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
hasmikharutyunyan: I thoght about oter word to describe this two states and found `ActivityStatus` do you have another suggestion?

mingazhev: deactivate and activate as different verbs was ok for me

Api/AdministratorServices/CounterpartyManagementService.cs
mingazhev: If it is an audit log, it should come in same transaction. Both should succeed or none

Api/Models/Management/Enums/ActivityStatus.cs
```diff
@@ -0,0 +1,8 @@
+namespace HappyTravel.Edo.Api.Models.Management.Enums
+{
+    public enum ActivityStatus
```
mingazhev, line 3: Do we need an enum for two statuses?

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
hasmikharutyunyan: So, you think is better to have two api s for that?

mingazhev: Yes, I think it will be better

Api/Models/Management/Enums/ActivityStatus.cs
hasmikharutyunyan: It is for api to indicate status

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
hasmikharutyunyan: I thought and about that and prefer to have one api, to not repeat the logic

kirillta: The logic remains the same, just two endpoints leading to one private method, though

kirillta (approved):

kirillta (approved):

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
```diff
@@ -119,17 +119,38 @@ public async Task<IActionResult> UpdateCounterparty(int counterpartyId, [FromBod
...
+            return NoContent();
+        }
+        
+        /// <summary>
+        ///  Activates specified counterparty
+        /// </summary>
+        /// <param name="counterpartyId">Id of the counterparty.</param>
+        /// <param name="request">Request data for Activation.</param>
+        /// <returns></returns>
+        [HttpPut("{counterpartyId}/activate")]
```
mingazhev, line 147: I think it'll better to use POST for such requests.
We already have a verb "activate" and it's better to use POST for actions that do not fit REST verbs

Api/Models/Management/ActivityStatusChangeRequest.cs
```diff
@@ -0,0 +1,14 @@
+using HappyTravel.Edo.Api.Models.Management.Enums;
+
+namespace HappyTravel.Edo.Api.Models.Management
+{
+    public readonly struct ActivityStatusChangeRequest
+    {
+
```
mingazhev, line 7: Redundant blank line

Api/Models/Management/ActivityStatusChangeRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+        public ActivityStatusChangeRequest(string reason)
```
mingazhev, line 8: Shouldn't  [jsonconstructor] be present there?

Api/Models/Management/AuditEvents/CounterpartActivityStatusChangeEventData.cs
```diff
@@ -0,0 +1,16 @@
+namespace HappyTravel.Edo.Api.Models.Management.AuditEvents
+{
+    public readonly struct CounterpartActivityStatusChangeEventData
```
mingazhev, line 3: Conterpart**y**

HappyTravel.Edo.Common/Enums/ManagementEventType.cs
```diff
@@ -5,5 +5,10 @@ public enum ManagementEventType
         None = 0,
         CounterpartyVerification = 10,
         AdministratorRegistration = 20,
+        CounterpartyDeactivation = 30,
+        AgencyDeactivation = 40,
+        CounterpartyActivation = 50,
+        AgencyActivation = 60
+
```
mingazhev, line 12: Redundant blank line

HappyTravel.Edo.UnitTests/Tests/AdministratorServices/CounterpartyManagementServiceTests/ActivationTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            var (_, isFailure, error) = await counterpartyManagementService.ActivateAgency(3, "Test reason");
+
+            var counterparty = context.Counterparties.Single(c => c.Id == 2);
+            Assert.False(isFailure);
+            Assert.True(counterparty.IsActive);
+        }
+
+
+        private readonly AdministratorServicesMockCreationHelper _mockCreationHelper;
```
mingazhev, line 203: Why do we need a shareable field there? Looks like moving its creation to method will be better

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
hasmikharutyunyan: I think POST is better for creation

HappyTravel.Edo.UnitTests/Tests/AdministratorServices/CounterpartyManagementServiceTests/ActivationTests.cs
hasmikharutyunyan: You mean,  have data here not in separate file ?

Raibo (approved):

Api/Models/Management/ActivityStatusChangeRequest.cs
hasmikharutyunyan: It is work without that attribute

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
mingazhev: Yes, PUT is good for idemponent, but this request is not idemponent. End in general - PUT is to update something, that's how it is in wiki:
`Store the representation in the request body as the (new) state of the resource. Subsequent GET access on the resource is expected to return this representation with HTTP status 200 until PATCHed, a different representation is PUT, or the state of the resource is DELETEd. `

Do we have new state in the request body?

HappyTravel.Edo.UnitTests/Tests/AdministratorServices/CounterpartyManagementServiceTests/ActivationTests.cs
mingazhev: I mean two things:
1. These two lines are repeated in each test and do not describe anything useful:
```
var context = _mockCreationHelper.GetContextMock().Object;
            var counterpartyManagementService = _mockCreationHelper.GetCounterpartyManagementService(context);
```

this could be moved to a method at least

2. Looks like there is no state that is needed to be shared. Creating and instantiating it in counstructor is worse than getting in method.

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
mingazhev: On the other hand, POST is used like "generic" or "multipurpose" request

HappyTravel.Edo.UnitTests/Tests/AdministratorServices/CounterpartyManagementServiceTests/ActivationTests.cs
hasmikharutyunyan: 1. In tests for checking result is used context, for example an.
I have done those for consistency with other tests, anyway i think it is not a big deal.

mingazhev (approved):
Do not merge for now, will deploy later

PR #284, opened 2020-11-10
Feature/NIJO-1030 add room supplier reference code to voucher

DFSko:


mingazhev (approved):

kirillta (changes requested):
It would be nice if you add tests to this class as well

DFSko:
> It would be nice if you add tests to this class as well

testing invoices and vouchers added in NIJO-1027 branch

kirillta (approved):

PR #285, opened 2020-11-11
Prevent HttpBasedAgentContextService from finding relations which are disabled

Raibo:


Raibo:
The change looks simple and theoretically have to work, but I will be able to test this only when NIJO-1017 will be done.

hasmikharutyunyan (approved):

mingazhev:
It's ok but please remove the  "IsAgent" property from Agent and make appropriate changes in the existing logic (where property is used), as we discussed

mingazhev (commented):
It's ok but please remove the "IsAgent" property from Agent and make appropriate changes in the existing logic (where property is used), as we discussed

mingazhev (approved):

mingazhev (approved):

PR #286, opened 2020-11-11
Feature/NIJO-1028 added additional properties to booking documents

DFSko:


Api/Extensions/MaskStringExtension.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Linq;
+
+namespace HappyTravel.Edo.Api.Extensions
+{
+    public static class MaskStringExtension
```
mingazhev, line 6: You should use Formatters.PersonNameFormatters instead of this

Api/Models/Payments/PaymentReceipt.cs
```diff
@@ -24,5 +34,44 @@ namespace HappyTravel.Edo.Api.Models.Payments
...
+
+        public readonly struct ReceiptItemInfo
+        {
+            [JsonConstructor]
+            public ReceiptItemInfo(DateTime? deadlineDate, RoomTypes roomType)
+            {
+                DeadlineDate = deadlineDate;
+                RoomType = roomType;
+            }
+
```
mingazhev, line 71: Just one blank line will be enough

Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
```diff
@@ -192,6 +208,11 @@ private async Task<Result<(DocumentRegistrationInfo RegistrationInfo, BookingInv
                 : Result.Success(lastInvoice);
         }
 
+        private static readonly HashSet<BookingStatuses> InvoiceCanceledBookingStatuses = new HashSet<BookingStatuses>
```
mingazhev, line 211: I think we should review documentation for consistency. We do not allow sending invoices for cancelled bookings.
Should we use invoice statuses - actual or cancelled?
I think we do not

mingazhev (changes requested):

Api/Extensions/MaskStringExtension.cs
DFSko: Formatters.PersonNameFormatters takes two strings (firstname and lastname separately), but MainPassengerName is one string

Api/Models/Payments/PaymentReceipt.cs
DFSko: Many other structures have two blank lines after constructor, including this class. Need to fix a whole class?

Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
DFSko: Status calculated on generating invoice and stored in Invoices table in json data. This hashset needed for calculating invoice status

Api/Models/Payments/PaymentReceipt.cs
DFSko: This line removed

Api/Extensions/MaskStringExtension.cs
mingazhev: It's better to add full main passenger name to the invoice (or maybe add already masked) and make sure that passenger name masking works identically in all cases

Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
mingazhev: I mean that we do not need to add such thing as invoice status, at least for now. I'll correct the documentation.

Api/Extensions/MaskStringExtension.cs
DFSko: fixed

mingazhev (approved):

PR #287, opened 2020-11-12
Feature/nijo 955 change data provider to supplier models

mingazhev:
Will break frontend

### DIFF
--- a/Api/Services/Locations/LocationService.cs
+++ b/Api/Services/Locations/LocationService.cs
@@ -145,7 +145,7 @@ namespace HappyTravel.Edo.Api.Services.Locations
                         : location.Name,
                     Source = location.Source,
                     Type = location.Type,
-                    DataProviders = location.DataProviders,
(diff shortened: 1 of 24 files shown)


kirillta (approved):

Api/Services/Locations/LocationService.cs
```diff
@@ -145,7 +145,7 @@ public async Task Set(IEnumerable<Models.Locations.Location> locations)
                         : location.Name,
                     Source = location.Source,
                     Type = location.Type,
-                    DataProviders = location.DataProviders,
+                    DataProviders = location.Suppliers,
```
hasmikharutyunyan, line 148: It si will be changed later?

hasmikharutyunyan (approved):

Api/Services/Locations/LocationService.cs
mingazhev: Yes, data that is stored in db will be changed in further PR-s

PR #288, opened 2020-11-13
Feature/nijo 1013 load contract file

Raibo:


Api/Services/Files/ContractFileService.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            return await ValidateFile()
+                .Bind(Upload);
+
+
+            Result ValidateFile() =>
+                Result.Success()
+                    .Ensure(() => file != null, "Couldn't get any file")
+                    .Ensure(() => file.Length > 0, "Got an empty file")
+                    .Ensure(() => Path.GetExtension(file?.FileName)?.ToLower() == ".pdf", "The file must have extension '.pdf'");
```
mingazhev, line 34: this should be extracted to constants

mingazhev, line 34: string.Equals with stringComparison.OrdinalIgnoreCase will be shorter

Api/Services/Files/ContractFileService.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                await using var stream = file.OpenReadStream();
+                return await _amazonS3ClientService.Add(_bucketName, GetKey(counterpartyId), stream, S3CannedACL.Private);
+            }
+        }
+
+
+        public async Task<Result<Stream>> Get(int counterpartyId)
+        {
+            var (_, isFailure, stream, _) = await _amazonS3ClientService.Get(_bucketName, GetKey(counterpartyId));
```
mingazhev, line 47: Why not in pipeline format? Will be better if top-level methods has the same style

Api/Services/Files/ContractFileService.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            if (isFailure)
+                return Result.Failure<Stream>("Couldn't get a contract file");
+
+            return Result.Success(stream);
```
mingazhev, line 52: looks like Result.Success could be omitted

Api/Controllers/AdministratorControllers/CounterpartiesController.cs
```diff
@@ -173,6 +177,43 @@ public async Task<IActionResult> GetCounterpartyPredictions(string query)
...
+        [ProducesResponseType(typeof(ProblemDetails), (int)HttpStatusCode.BadRequest)]
+        [AdministratorPermissions(AdministratorPermissions.CounterpartyManagement)]
+        public async Task<IActionResult> GetContractFile(int counterpartyId)
+        {
+            var (_, isFailure, stream, error) = await _contractFileService.Get(counterpartyId);
+
+            if (isFailure)
+                return BadRequest(ProblemDetailsBuilder.Build(error));
+
+            return File(stream, "application/pdf");
```
mingazhev, line 212: Is getting the strict contact type is responsibility of controller?
I think controller should not know about the file format, this logic is bound to a service

Api/Services/Files/ContractFileService.cs
Raibo: `string.Equals(Path.GetExtension(file?.FileName), PdfFileExtension, StringComparison.OrdinalIgnoreCase)`
vs
`Path.GetExtension(file?.FileName)?.ToLower() == PdfFileExtension`
Not shorter at all

Api/Services/Files/ContractFileService.cs
Raibo: There is simply nothing to put into local methods, we simply call the other service and swallow possible error, nothing more.

Api/Services/Files/ContractFileService.cs
mingazhev: ok, so use tolowerinvariant to discard the culture, this is a good practice

mingazhev (approved):
with a little comment

PR #294, opened 2020-11-17
Feature/nijo 1019 uploading images

Raibo:


Api/Services/Files/ImageFileService.cs
```diff
@@ -0,0 +1,189 @@
...
+            _s3FolderName = options.Value.S3FolderName;
+            _agentContextService = agentContextService;
+            _edoContext = edoContext;
+            _dateTimeProvider = dateTimeProvider;
+        }
+
+
+        public async Task<Result> Add(IFormFile file)
+        {
+            var agentContext = await _agentContextService.GetAgent();
```
mingazhev, line 42: You should pass agentContext from controller

Api/Services/Files/ImageFileService.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private string GetKey(int agencyId, string fileName) => $"{_s3FolderName}/{agencyId}/{fileName.ToLower()}";
+
+
+        private const int MinimumWidth = 200;
+        private const int MaximumWidth = 800;
+        private const int MinimumHeight = 200;
+        private const int MaximumHeight = 800;
+
+        private readonly List<string> _imageExtensions = new List<string>{"jpeg", "jpg"};
```
mingazhev, line 179: Please use hashset for such things

mingazhev: And this could be static readonly

Api/Services/Files/ImageFileService.cs
```diff
@@ -0,0 +1,189 @@
...
+            return await Validate()
+                .Bind(Upload)
+                .Map(StoreLink);
+
+
+            async Task<Result> Validate()
+            {
+                return await Result.Success()
+                    .Ensure(() => file != default, "Could not get the file")
+                    .Ensure(() => _imageExtensions.Contains(Path.GetExtension(file?.FileName)?.ToLower()), "The file must have extension of a jpeg image")
```
mingazhev, line 53: ToLowerInvariant

Api/Controllers/AgentControllers/AgencyImagesController.cs
```diff
@@ -0,0 +1,57 @@
...
+
+
+        /// <summary>
+        ///     Uploads an image
+        /// </summary>
+        [HttpPut("images")]
+        [ProducesResponseType((int) HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
+        [InAgencyPermissions(InAgencyPermissions.AgencyImagesManagement)]
+        public async Task<IActionResult> Add([FromForm] IFormFile file) => OkOrBadRequest(await _imageFileService.Add(file));
```
mingazhev, line 32: Don't think that we need such general-purpose endpoint.
For now we need a banner and logo only.
Why not loading and getting them throw separate endpoints?

@kirillta what do you think?

AlexeyShok: Why is the HttpPut request used? If the method "Add" adds a picture then it's should be a HttpPost. Either the method is named incorrectly.

Api/Services/Files/ImageFileService.cs
```diff
@@ -0,0 +1,189 @@
...
+                            $"Image width must be in range from {MinimumWidth} to {MaximumWidth}")
+                        .Ensure(() => imageInfo.ImageHeight >= MinimumHeight && imageInfo.ImageHeight <= MaximumHeight,
+                            $"Image height must be in range from {MinimumWidth} to {MaximumWidth}");
+                }
+            }
+
+
+            async Task<Result<string>> Upload()
+            {
+                await using var stream = file.OpenReadStream();
```
AlexeyShok, line 83: Problems may arise due to incorrect work with the file stream. The stream is opened for reading twice: during validation and then when loaded into S3. Possible slowdown or crash when rereading from the file stream. This solution needs to be tested thoroughly.

Api/Services/Files/ImageFileService.cs
kirillta: Why, isn't that an unnecessary dependency?

Api/Controllers/AgentControllers/AgencyImagesController.cs
```diff
@@ -0,0 +1,57 @@
...
+
+
+        /// <summary>
+        ///     Gets all uploaded images for the current agency
+        /// </summary>
+        [HttpGet("images")]
+        [ProducesResponseType(typeof(List<SlimUploadedImage>), (int) HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
```
kirillta, line 40: This method returns only 200

Api/Models/Agencies/SlimUploadedImage.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// File name of the image
+        /// </summary>
+        public string FileName { get; }
+
+        /// <summary>
+        /// Download link of the image
+        /// </summary>
+        public string Link { get; }
```
kirillta, line 25: If I remember right we name it `Url` in other places @AlexeyShok

kirillta (changes requested):

Api/Models/Agencies/SlimUploadedImage.cs
AlexeyShok: Yes, we named it Url in hiroshima.

Api/Controllers/AgentControllers/AgencyImagesController.cs
Raibo: @AlexeyShok , HttpPut is used due to the fact that new picture with the same name replaces the old one.
If you think that method named incorrectly, suggest a better name.

Api/Services/Files/ImageFileService.cs
Raibo: In previous review I was told to move agentContext from controller to service. 
Here I am told to move it from service to controller.
This is strange.

Api/Services/Files/ImageFileService.cs
Raibo: Changed to save bytes and memory stream

Api/Models/Agencies/SlimUploadedImage.cs
Raibo: Renamed

Api/Controllers/AgentControllers/AgencyImagesController.cs
Raibo: @mingazhev Why not. So, `.../banner` and `.../logo` PUT/GET endpoints?

Api/Services/Files/ImageFileService.cs
mingazhev: I told so? Maybe I was mistaken

Api/Controllers/AgentControllers/AgencyImagesController.cs
mingazhev: Yes, separate endpoints will be better. Del will be good too

Api/Services/Files/ImageFileService.cs
mingazhev: Could you give a link?

mingazhev: And Kirill is right, could not find where this dependency used now

mingazhev (commented):
Need to change the endpoints logic

Api/Services/Files/ImageFileService.cs
Raibo: Here is the link
https://github.com/happy-travel/edo/pull/283#discussion_r520647510

Api/Controllers/AgentControllers/AgencyImagesController.cs
AlexeyShok: If HttpPut remains, then I would call method AddOrReplace or Replace, but not Add

Api/Controllers/AgentControllers/AgencyImagesController.cs
```diff
@@ -0,0 +1,115 @@
...
+
+
+        /// <summary>
+        ///     Uploads an image for banner
+        /// </summary>
+        [HttpPut("banner")]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int)HttpStatusCode.BadRequest)]
+        [InAgencyPermissions(InAgencyPermissions.AgencyImagesManagement)]
+        public async Task<IActionResult> AddOrReplaceBanner([FromForm] IFormFile file)
```
mingazhev, line 35: SetBanner maybe?

mingazhev (approved):
With a small comment

kirillta (approved):

Api/Controllers/AgentControllers/AgencyImagesController.cs
Raibo: I also considered AddOrReplace, but after discussing with Vadim, renamed to Set

PR #310, opened 2020-11-23
Add endpoint for agents to get contract files (NIJO-1014)

Raibo:


Api/Services/Files/IContractFileService.cs
```diff
@@ -10,6 +11,8 @@ public interface IContractFileService
     {
         Task<Result> Add(int counterpartyId, IFormFile file);
 
+        Task<Result<(Stream stream, string contentType)>> GetForAgent(AgentContext agentContext);
```
mingazhev, line 14: This should be in a separate service, as we've discussed: agent and admin services should be separated

HappyTravel.Edo.Data/Migrations/20201123095157_ChangeAllPrivilegesValue.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace HappyTravel.Edo.Data.Migrations
+{
+    public partial class ChangeAllPrivilegesValue : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(
+                "UPDATE \"AgentAgencyRelations\" SET \"InAgencyPermissions\" = 2147483646 WHERE \"InAgencyPermissions\" in (16382, 16383, 32766);"
```
mingazhev, line 10: Why 16383?
And maybe split this to two revertable migrations?

mingazhev (changes requested):

HappyTravel.Edo.Data/Migrations/20201123095157_ChangeAllPrivilegesValue.cs
Raibo: 16383 is because i found this value in master branch in InAgencyPermissions file. So, just in case.
Multiple values is because I don't know when exactly this migration will be executed compared to other migrations which also alters this table.

I don't have idea how could we split this to two migrations. What exactly do you mean?

Api/Services/Files/IContractFileService.cs
Raibo: Moved to a separate service

mingazhev (approved):

PR #311, opened 2020-11-23
Feature/nijo 957 currency buffer removed

kirillta:


Api/Services/CurrencyConversion/CurrencyConverterService.cs
```diff
@@ -23,34 +25,34 @@ public CurrencyConverterService(ICurrencyRateService rateService)
...
             
             return Result.Success(convertedDetails);
         }
 
 
-        private const decimal ConversionBuffer = (decimal)0.005;
-        
         // Only USD Currency is supported for now.
-        private const Currencies TargetCurrency = Currencies.USD;
+        public const Currencies TargetCurrency = Currencies.USD;
```
mingazhev, line 53: Why public?

kirillta: In that case — tests, but I do not see any issues making this field public

mingazhev (approved):

PR #315, opened 2020-11-24
Fixed notifications to come only after booking confirmed

mingazhev:


Api/Services/Company/CompanyService.cs
```diff
@@ -1,37 +1,45 @@
...
-            return Result.Success(_companyInfo);
+            return info ?? Result.Failure<CompanyInfo>("Could not find company information");
         }
 
-
+        private static readonly TimeSpan CompanyInfoCacheLifeTime = TimeSpan.FromHours(1); 
+        
         private readonly EdoContext _context;
-        private static CompanyInfo _companyInfo;
+        private readonly IDoubleFlow _flow;
```
hasmikharutyunyan, line 43: why changed to redis? It is not a large information

mingazhev: I though about multi instances, it will be faster for them to fetch data from redis, not from db.
Do you think there should be memory flow?

mingazhev: I've changed previous implementation because it was not thread safe. There were a static property that could be changed from different threads in a concurrent manner, which could cause hardly reproduceable bugs

hasmikharutyunyan: No, I mean it is previously saved in static field and fetched from db only once (I think It will not disturb to have  multi instances)

hasmikharutyunyan (approved):

kirillta (approved):

PR #316, opened 2020-11-24
Technical/nijo 1075 remove agency from routes

mingazhev:


Api/Controllers/AgentControllers/CounterpartiesController.cs
```diff
@@ -14,7 +14,7 @@ namespace HappyTravel.Edo.Api.Controllers.AgentControllers
 {
     [ApiController]
     [ApiVersion("1.0")]
-    [Route("api/{v:apiVersion}/counterparties")]
+    [Route("api/{v:apiVersion}/counterparty")]
```
hasmikharutyunyan, line 17: why counterparty?

mingazhev: Because an agent can be affiliated only with one counterparty. And he or she gets counterparty not as a resource
/counterparties/{id}, but as a resource names "counterparty", where current affiliated counterparty is

kirillta: As for me, that's confusing

hasmikharutyunyan (approved):

PR #317, opened 2020-11-24
Get bookings charged one day before deadline and at finalization (NIJO-819)

Raibo:


Api/Services/Accommodations/Bookings/BookingRegistrationService.cs
```diff
@@ -246,7 +264,9 @@ async Task<string> RegisterBooking(BookingAvailabilityInfo bookingAvailability)
 
             async Task<Result<Data.Booking.Booking, ProblemDetails>> PayUsingAccountIfDeadlinePassed(Data.Booking.Booking bookingInPipeline)
             {
-                if (!availabilityDeadline.HasValue || availabilityDeadline > _dateTimeProvider.UtcNow())
+                var daysBeforeDeadline = Infrastructure.Constants.Common.DaysBeforeDeadlineWhenPayForBooking;
+
+                if (!availabilityDeadline.HasValue || availabilityDeadline > _dateTimeProvider.UtcNow().AddDays(daysBeforeDeadline))
```
mingazhev, line 269: Should we check checkin date there?

HappyTravel.Edo.UnitTests/Tests/Services/Accommodations/Bookings/BookingsProcessingServiceTests/Charging/GettingForCharging.cs
```diff
@@ -18,13 +18,15 @@ public class GettingForCharging
...
             {
-                CreateBooking(id: 1, deadlineDate: new DateTime(2021, 12, 1)),
-                CreateBooking(id: 2, deadlineDate: new DateTime(2021, 12, 2)),
-                CreateBooking(id: 3, deadlineDate: new DateTime(2021, 12, 3)),
-                CreateBooking(id: 4, deadlineDate: new DateTime(2021, 12, 4)),
+                CreateBooking(id: 1, deadlineDate: new DateTime(2021, 12, 1).AddDays(daysBeforeDeadlineToPay)),
+                CreateBooking(id: 2, deadlineDate: new DateTime(2021, 12, 2).AddDays(daysBeforeDeadlineToPay)),
+                CreateBooking(id: 3, deadlineDate: new DateTime(2021, 12, 3).AddDays(daysBeforeDeadlineToPay)),
+                CreateBooking(id: 4, deadlineDate: new DateTime(2021, 12, 4).AddDays(daysBeforeDeadlineToPay)),
                 CreateBooking(id: 5, deadlineDate: null)
```
mingazhev, line 30: You should change the data, not using any logic

Api/Services/Accommodations/Bookings/BookingsProcessingService.cs
```diff
@@ -38,9 +38,12 @@ public class BookingsProcessingService : IBookingsProcessingService
         public Task<List<int>> GetForCapture(DateTime date)
         {
             date = date.Date;
+            var daysBeforeDeadline = Infrastructure.Constants.Common.DaysBeforeDeadlineWhenPayForBooking;
+
             return _context.Bookings
                 .Where(IsBookingValidForCapturePredicate)
-                .Where(b => b.CheckInDate <= date || (b.DeadlineDate.HasValue && b.DeadlineDate.Value.Date <= date))
+                .Where(b => b.CheckInDate <= date.AddDays(daysBeforeDeadline) 
+                    || (b.DeadlineDate.HasValue && b.DeadlineDate.Value.Date <= date.AddDays(daysBeforeDeadline)))
```
mingazhev, line 46: Should be there add days? Not substract?

mingazhev (changes requested):

HappyTravel.Edo.UnitTests/Tests/Services/Accommodations/Bookings/BookingsProcessingServiceTests/Charging/GettingForCharging.cs
Raibo: This will lead to necessity to change tests every time the constant changes. 
I don't know if it's a good thing. What do you think?

Api/Services/Accommodations/Bookings/BookingsProcessingService.cs
Raibo: So, "today" might be 1st January. To make things happen a day earlier, we have to check as if the present day would be 2nd January, like "would it trigger tomorrow"
This logic seems to me easier to get than moving deadline dates.

Api/Services/Accommodations/Bookings/BookingRegistrationService.cs
Raibo: I thought that absence of deadline date means that booking can be refunded even after check in date, hence there are no politics that could cut refundable amount.
This means that we don't need to capture immediately in this case and can leave it for PaymentProcessing to take care about capturing.

mingazhev: No, no booking can be refunded after checkin date.
There could be a whole, when we allow the agent to book after check in without any money on his account

Api/Services/Accommodations/Bookings/BookingsProcessingService.cs
mingazhev: Sounds ok. But I think you should calculate the value before passing it to `Where` clause. Does it compile to SQL in this way?

HappyTravel.Edo.UnitTests/Tests/Services/Accommodations/Bookings/BookingsProcessingServiceTests/Charging/GettingForCharging.cs
mingazhev: Sure, this constant is a part of the logic. When logic will change, tests will fail, it's Ok.
Then, when it will be a configurable value, we'll mock this value in different ways and check with raw test data again.
This tests breaks incapsulation (test starts to know that there is some constant and we should add this to our dates and etc.) and duplicating the method under tests logic.

mingazhev (commented):
Some comments

Api/Services/Accommodations/Bookings/BookingsProcessingService.cs
Raibo:
```      Executing DbCommand [Parameters=[@__AddDays_0='?' (DbType = DateTime)], CommandType='Text', CommandTimeout='30']
      SELECT b."Id"
      FROM "Bookings" AS b
      WHERE ((b."Status" IN (2, 3, 0, 1) AND b."PaymentMethod" IN (2)) AND (b."PaymentStatus" = 1)) AND ((b."CheckInDate" <= @__AddDays_0) OR ((b."DeadlineDate" IS NOT NULL) AND (DATE_TRUNC('day', b."DeadlineDate") <= @__AddDays_0)))```

Api/Services/Accommodations/Bookings/BookingRegistrationService.cs
Raibo: Made identical with GetForCharge / GetForCapture.

HappyTravel.Edo.UnitTests/Tests/Services/Accommodations/Bookings/BookingsProcessingServiceTests/Charging/GettingForCharging.cs
Raibo: Ok

mingazhev (approved):

PR #322, opened 2020-11-26
Technical/nijo 777 remove default values from flags

DFSko:


HappyTravel.Edo.Common/Enums/InAgencyPermissions.cs
```diff
@@ -8,22 +8,21 @@ namespace HappyTravel.Edo.Common.Enums
...
-        ObserveBalance = 128,
-        ObservePaymentHistory = 256,
-        AgencyToChildTransfer = 512,
-        ReceiveBookingSummary = 1024,
-        AgencyBookingsManagement = 2048,
-        AgentStatusManagement = 4096,
-        ObserveAgencyInvitations = 8192,
-        AgencyImagesManagement = 16384,
-        ObserveCounterpartyContract = 32768,
+        AgentInvitation = 1,
```
mingazhev, line 11: Is there a migration to change this in db?
And please add a comment for this enum, something about that database migration may be needed after enum changes (to not forget it in future)

mingazhev (changes requested):

HappyTravel.Edo.Common/Enums/InAgencyPermissions.cs
DFSko: Fixed

HappyTravel.Edo.Data/Migrations/20201126102757_RemoveDefaultValueFromInAgencyPermissions.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace HappyTravel.Edo.Data.Migrations
+{
+    public partial class RemoveDefaultValueFromInAgencyPermissions : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
mingazhev, line 8: This migration will update permissions only for agents that have only one permission. But we store flags combinations instead of separate values

mingazhev (changes requested):

mingazhev (changes requested):
Migration looks incorrect

mingazhev (approved):
looks fine

PR #324, opened 2020-11-26
Add logo and banner urls into vouchers (NIJO-1020)

Raibo:


Api/Services/Accommodations/Bookings/BookingDocumentsService.cs
```diff
@@ -54,6 +57,9 @@ public async Task<Result<BookingVoucherData>> GenerateVoucher(int bookingId, str
             if(isAccommodationFailure)
                 return Result.Failure<BookingVoucherData>(accommodationError.Detail);
 
+            var (isBannerSuccess, _, bannerImage, _) = await _imageFileService.GetBanner(agent);
```
mingazhev, line 60: Isn't it better to use maybe there? Semantically it does not look like a failure result

mingazhev (approved):

PR #333, opened 2020-12-02
added harcode for markup disable

hasmikharutyunyan:


Api/Services/Agents/CounterpartyService.cs
```diff
@@ -25,6 +27,9 @@ public class CounterpartyService : ICounterpartyService
 
         public async Task<Result<CounterpartyInfo>> Add(CounterpartyEditRequest counterparty)
         {
+            // TODO: Remove
```
mingazhev, line 30: Please add the tech debt task for removal

mingazhev: And this could be closer to usage

Api/Services/Agents/CounterpartyService.cs
```diff
@@ -59,8 +64,24 @@ public async Task<Result<CounterpartyInfo>> Add(CounterpartyEditRequest counterp
                 ParentId = null,
             };
             _context.Agencies.Add(defaultAgency);
-
             await _context.SaveChangesAsync();
+
+            //TODO: Remove
```
mingazhev, line 69: Need a task and longer description.
Something like "Temporary solution for CIS countries, registered before date"

kirillta:
The function is needed on Production tomorrow morning, but this PR will not on Production tomorrow morning. Pointless

PR #334, opened 2020-12-03
fixed empty error on cancellation booking endpoint

DFSko:
Maybe so conceived, but it looks like a bug


mingazhev (approved):
Sure, it is a bug

PR #335, opened 2020-12-03
implement agent markup bonuses

DFSko:


Api/Services/CurrencyConversion/CurrencyConverterService.cs
```diff
@@ -49,10 +54,28 @@ public CurrencyConverterService(ICurrencyRateService rateService, ICurrencyConve
         }
 
 
+        public async Task<Result<decimal>> Convert(Currencies source, Currencies target, decimal value)
+        {
+            if (source == target)
+                return value;
+
+            using var client = _clientFactory.CreateClient(HttpClientNames.CurrencyService);
```
kirillta, line 62: Again we use conversion by the library as in the method above, because it handles buffer values

Api/Services/Markups/PaymentMarkupService.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            Task<Result<PaymentMarkupDataWithValue>> GetValue(PaymentMarkupData data)
+                => GetMarkupValue(data, isWriteOff);
+        }
+
+
+        private async Task<Result<PaymentMarkupDataWithValue>> GetMarkupValue(PaymentMarkupData data, bool isWriteOff)
+        {
+            // TODO: get markup value
+            const decimal markup = 0;
```
kirillta, line 46: Hmmm

PR #336, opened 2020-12-03
Fixed agent and agency routes to be more consistent and restful

mingazhev:


Api/Controllers/AgentControllers/AgentsController.cs
```diff
@@ -404,7 +404,7 @@ private async Task<string> GetUserEmail()
         /// <summary>
         ///     Disables a given agent to operate using a given agency
         /// </summary>
-        [HttpPost("agents/{agentId}/disable")]
+        [HttpPost("agency/agents/{agentId}/disable")]
```
hasmikharutyunyan, line 407: why we need `agency`  here ? It is an agentController and I think it is ok to not use `ageny`

hasmikharutyunyan (commented):
maybe needed to change `AgentsController` to `AgentController`?

Api/Controllers/AgentControllers/AgentsController.cs
hasmikharutyunyan: Or replace this action into `AgencyController`

mingazhev: Maybe, I did not move methods there, tried to change only the routes to make code review easier.
I'm agree to move agency methods to the separate controllers, but I think routes will be the same

mingazhev:
> 
> 
> maybe needed to change `AgentsController` to `AgentController`?

Sure, I'll change later. Just tried to make PR as lightweight as possible

Api/Controllers/AgentControllers/AgentsController.cs
hasmikharutyunyan: So, It is will be done in the next PR-s, I can approve this?

mingazhev: Yes, if you are agree with the routes

hasmikharutyunyan (approved):

alexander-plekhanov (approved):

PR #340, opened 2020-12-04
fixed confirm payment always return booking not found

DFSko:


Api/Services/Accommodations/Bookings/CreditCardPaymentConfirmationService.cs
```diff
@@ -32,13 +32,23 @@ public async Task<Result> Confirm(int bookingId)
...
+                    select new {booking, confirmation.ConfirmedAt};
 
                 var data = await query.SingleOrDefaultAsync();
 
-                return data ?? Result.Failure<Booking>($"Booking with Id {bookingId} not found");
+                if(data is null)
+                    return Result.Failure<Booking>($"Booking with Id {bookingId} not found");
+
+                if(data.ConfirmedAt != default)
+                    return Result.Failure<Booking>("Payment already confirmed");
```
mingazhev, line 46: This should not be a failure. Could you return success result there?

mingazhev (approved):

PR #343, opened 2020-12-04
added OData support for agents and agencies booking

DFSko:


Api/Services/Accommodations/Bookings/BookingRecordsManager.cs
```diff
@@ -130,15 +130,37 @@ static List<BookedRoom> UpdateSupplierReferenceCodes(List<BookedRoom> existingRo
...
+                    },
+                    Data = new SlimAccommodationBookingInfo
+                    {
+                        Id = booking.Id,
+                        ReferenceCode = booking.ReferenceCode,
+                        AccommodationName = booking.AccommodationName,
+                        CountryName = booking.Location.Country,
+                        LocalityName = booking.Location.Locality,
+                        Deadline = booking.DeadlineDate,
+                        Price = new MoneyAmount(booking.TotalPrice, booking.Currency),
```
mingazhev, line 155: Would the price filter and sort work in this case?
Maybe we should use primitive types here

mingazhev (approved):
Looks fine for me

Api/Services/Accommodations/Bookings/BookingRecordsManager.cs
DFSko: Yes, sorting not work with this type, replaced to primitives

mingazhev (approved):

mingazhev:
Need to discuss model changes with frontend

mingazhev (approved):
Looks ok, let's wait for @kirillta

Api/Models/Bookings/SlimAccommodationBookingInfo.cs
```diff
@@ -22,40 +18,41 @@ public SlimAccommodationBookingInfo(Booking bookingInfo)
...
+        public int Id { get; init; }
+
+        public string ReferenceCode { get; init; }
 
-        public BookingStatuses Status { get; }
+        public BookingStatuses Status { get; init; }
+        public decimal Price { get; init; }
 
-        public MoneyAmount Price { get; }
+        public Currencies Currency { get; init; }
```
kirillta, line 38: Why? We agreed to use `MoneyAmount` across the system

DFSko: Because sorting and filtering not working on complex types

PR #344, opened 2020-12-06
added OData support for agents list

DFSko:


mingazhev (approved):
Looks fine, if we temorary remove the markup string from the answer

kirillta (approved):

PR #355, opened 2020-12-09
added checking for IsResent before sending new invitation

DFSko:


Api/Services/Agents/AgentInvitationService.cs
```diff
@@ -83,7 +83,13 @@ async Task<Result<AgentInvitation>> GetExistingInvitation()
                     .AgentInvitations
                     .SingleOrDefaultAsync(i => i.CodeHash == invitationCode);
 
-                return invitation ?? Result.Failure<AgentInvitation>($"Invitation with Code {invitationCode} not found");
+                if(invitation is null)
```
mingazhev, line 86: formatting

mingazhev (changes requested):

Api/Services/Agents/AgentInvitationService.cs
DFSko: Fixed

mingazhev (approved):

PR #371, opened 2020-12-15
Saving intermediate booking info to cache instead of DB

mingazhev:


Api/Services/Accommodations/Bookings/BookingRegistrationService.cs
```diff
@@ -128,7 +130,14 @@ async Task<string> Register(BookingAvailabilityInfo bookingAvailability)
                 .Bind(GetAccommodationBookingInfo)
                 .Finally(WriteLog);
 
+            
+            Task<Result<(AccommodationBookingRequest, string), ProblemDetails>> GetCachedRequestInfo(string referenceCode) => _requestStorage.Get(referenceCode).ToResultWithProblemDetails();
```
kirillta, line 134: The method still names after cache

kirillta (approved):

Api/Services/Accommodations/Bookings/BookingRegistrationService.cs
mingazhev: Fixed

PR #375, opened 2020-12-17
removed preferred currency from registration

DFSko:


Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -119,7 +120,7 @@ async Task<Result<CounterpartyInfo>> UpdateCounterparty(Counterparty counterpart
                 counterpartyToUpdate.Phone = changedCounterpartyInfo.Phone;
                 counterpartyToUpdate.Website = changedCounterpartyInfo.Website;
                 counterpartyToUpdate.PostalCode = changedCounterpartyInfo.PostalCode;
-                counterpartyToUpdate.PreferredCurrency = changedCounterpartyInfo.PreferredCurrency;
+                counterpartyToUpdate.PreferredCurrency = Currencies.USD;
```
mingazhev, line 123: Please add a comment that this was hardcoded because the only current supporting currency is USD

mingazhev (approved):
With little comment

Api/AdministratorServices/CounterpartyManagementService.cs
DFSko: Fixed

PR #383, opened 2020-12-21
Generating invoice only for confirmed booking

mingazhev:
Invoice generated and sent only after booking confirmation

### DIFF
--- a/Api/Services/Accommodations/Bookings/BookingChangesProcessor.cs
+++ b/Api/Services/Accommodations/Bookings/BookingChangesProcessor.cs
@@ -97,7 +97,7 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Bookings
             await GetBookingInfo(booking.ReferenceCode, booking.LanguageCode)
                 .Tap(Confirm)
                 .Tap(NotifyBookingFinalization)
-                .Check(GenerateInvoice)
+                .Bind(GenerateInvoice)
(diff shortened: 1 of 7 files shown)


Api/Services/Accommodations/Bookings/BookingChangesProcessor.cs
```diff
@@ -95,6 +97,7 @@ public async Task ProcessConfirmation(Booking booking, EdoContracts.Accommodatio
             await GetBookingInfo(booking.ReferenceCode, booking.LanguageCode)
                 .Tap(Confirm)
                 .Tap(NotifyBookingFinalization)
+                .Check(GenerateInvoice)
```
hasmikharutyunyan, line 100: maybe another name ? `CaneGenerateInvoec`, `GenerateIfCan` or something better ?

mingazhev: This is the issue with the method name discussed there:
https://github.com/vkhorikov/CSharpFunctionalExtensions/issues/242

Check is bad name, does not cover all usages.
In this case there is an action with result, I'll change for Bind now to be more clear

mingazhev: Changed to `Bind`

hasmikharutyunyan (approved):

kirillta (approved):

PR #385, opened 2020-12-21
Splitted two booking flows to separate services

mingazhev:
Credit card and account flows differ much and do not have real duplicate code so I've splitted them to separate small services.
+ in addition I've removed capturing credit card payment if deadline passed - we've authorized money before and I don't think that it is vital to capture at the same time.
This is discussable


Api/Services/Accommodations/Bookings/Flows/ICreditCardBookingFlow.cs
```diff
@@ -0,0 +1,14 @@
+using System.Threading.Tasks;
+using CSharpFunctionalExtensions;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Api.Models.Bookings;
+
+namespace HappyTravel.Edo.Api.Services.Accommodations.Bookings.Flows
+{
+    public interface ICreditCardBookingFlow
```
kirillta, line 8: Both class names are ambiguous. What about `IBankCardBookingFlow` and `IFinancialAccountBookingFlow`?

kirillta:
> * in addition I've removed capturing credit card payment if deadline passed - we've authorized money before and I don't think that it is vital to capture at the same time

That's OK while one cannot cancel such bookings

Api/Services/Accommodations/Bookings/Flows/ICreditCardBookingFlow.cs
mingazhev: Renamed

mingazhev:
> 
> 
> > * in addition I've removed capturing credit card payment if deadline passed - we've authorized money before and I don't think that it is vital to capture at the same time
> 
> That's OK while one cannot cancel such bookings

Good note, reverted money capture, at least for now

kirillta (approved):

PR #403, opened 2020-12-28
Fixed attached entity twice error

mingazhev:


Api/Services/Accommodations/Bookings/Payments/BookingPaymentInfoService.cs
```diff
@@ -105,17 +106,21 @@ public async Task<Result<(int AgentId, int AgencyId)>> GetServiceBuyer(string re
         }
 
 
-        public async Task<Result<AgencyAccount>> GetChargingAccount(string referenceCode)
+        public async Task<Result<int>> GetChargingAccountId(string referenceCode)
         {
             var (_, isFailure, booking, error) = await _bookingRecordsManager.Get(referenceCode);
             if (isFailure)
-                return Result.Failure<AgencyAccount>(error);
+                return Result.Failure<int>(error);
```
kirillta, line 113: Why?

mingazhev: We don't need the whole AgencyAccount entity outside, account id is enough

kirillta (approved):

DFSko (approved):

PR #406, opened 2020-12-29
added missing properties to payment receipt

DFSko:


Api/Services/Payments/PaymentNotificationService.cs
```diff
@@ -34,7 +36,23 @@ public Task<Result> SendReceiptToCustomer((DocumentRegistrationInfo Registration
                 Method = EnumFormatters.FromDescription(paymentReceipt.Method),
                 InvoiceNumber = paymentReceipt.InvoiceInfo.Number,
                 InvoiceDate = FormatDate(paymentReceipt.InvoiceInfo.Date),
-                ReferenceCode = paymentReceipt.ReferenceCode
+                ReferenceCode = paymentReceipt.ReferenceCode,
+                AccommodationName = paymentReceipt.AccommodationName,
+                RoomTypes = paymentReceipt.ReceiptItems.Select(r => new PaymentReceiptData.RoomDetails
```
mingazhev, line 41: Please use RoomDetails as it is done in voucher. I've changed the documentation

Api/Services/Payments/PaymentNotificationService.cs
```diff
@@ -34,7 +36,23 @@ public Task<Result> SendReceiptToCustomer((DocumentRegistrationInfo Registration
                 Method = EnumFormatters.FromDescription(paymentReceipt.Method),
                 InvoiceNumber = paymentReceipt.InvoiceInfo.Number,
```
mingazhev, line 37: Please move this method to BookingMailingService, this service looks like having incorrect name.
We'll split BookingMailingService to separate service later

DFSko: Fixed

Api/Services/Payments/PaymentNotificationService.cs
DFSko: Fixed

mingazhev (approved):

PR #408, opened 2020-12-30
Simplified async responses processing

mingazhev:
@oleg1std could you additionally check that nothing vital was not deleted?


kirillta (approved):

oleg1std (approved):

oleg1std:
> @oleg1std could you additionally check that nothing vital was not deleted?

looks valid

PR #412, opened 2021-01-04
fixed getting booking by agent with permission

DFSko:


mingazhev:
Will do in another way, will use this branch later

PR #414, opened 2021-01-11
added language usage to analytics

DFSko:


Api/Controllers/AgentControllers/AvailabilitiesController.cs
```diff
@@ -55,9 +56,11 @@ public class AvailabilitiesController : BaseController
         public async Task<IActionResult> StartAvailabilitySearch([FromBody] AvailabilityRequest request)
         {
             Counters.WideAccommodationAvailabilitySearchTimes.Inc();
+            var rcf = Request.HttpContext.Features.Get<IRequestCultureFeature>();
```
mingazhev, line 59: Should not we use existing LanguageCode there? @kirillta

DFSko: fixed

kirillta (approved):

mingazhev (approved):

PR #418, opened 2021-01-12
Feature/has contract flag

mingazhev:
Added a flag "HasContract" to allow frontend show a button "Download contract" or not to show if contract is not available


Api/Services/Files/ContractFileManagementService.cs
```diff
@@ -39,6 +42,19 @@ async Task<Result> Upload()
...
+
+            async Task<Result> UpdateCounterparty()
+            {
+                var counterparty = await _context.Counterparties.FindAsync(counterpartyId);
+                if (counterparty is null)
+                    return Result.Failure($"Could not find counterparty with id {counterpartyId}");
+
+                counterparty.HasContract = true;
+                _context.Update(counterparty);
+                await _context.SaveChangesAsync();
```
DFSko, line 55: Not need to update all properties in conterpary

mingazhev: Fixed

DFSko (approved):

Api/Services/Files/ContractFileManagementService.cs
hasmikharutyunyan: As I remember we have decided to do this kind of updates when it is affect on speed, not everywhere.

mingazhev: Agree, I've changed because I don't think that it is very important

Api/Models/Agents/CounterpartyInfo.cs
```diff
@@ -106,5 +99,10 @@ namespace HappyTravel.Edo.Api.Models.Agents
         /// E-mail for billing operations
         /// </summary>
         public string BillingEmail { get; }
+
+        /// <summary>
+        /// True if contract is loaded to counterparty
```
kirillta, line 104: What contract? The contract with HT? `IsContractConcluded` or `IsContractUploaded` maybe?

HappyTravel.Edo.Data/Agents/Counterparty.cs
```diff
@@ -26,5 +26,6 @@ public class Counterparty
         public string BillingEmail { get; set; }
         public string VatNumber { get; set; }
         public bool IsActive { get; set; }
+        public bool HasContract { get; set; }
```
kirillta, line 29: And here as well

Api/Models/Agents/CounterpartyInfo.cs
mingazhev: Yes, the contract with HT.
Changed to IsContractUploaded

HappyTravel.Edo.Data/Agents/Counterparty.cs
mingazhev: Fixed too

kirillta (approved):

PR #440, opened 2021-01-25
Added api client management

mingazhev:
The idea of 1-st version API client is to bind some service client to an agent.
And all that this service client will do, will looks the same as this agent do.

Later we'll find more valid approach, current one is just the fastest to implement


Api/Controllers/AdministratorControllers/AgentsController.cs
```diff
@@ -86,7 +88,48 @@ public async Task<IActionResult> MoveAgentToAgency([FromRoute] int agentId, [Fro
         }
         
         
+        /// <summary>
+        /// Move agent from one agency to another
```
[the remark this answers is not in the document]
mingazhev: Fixed, thanks

hasmikharutyunyan (approved):

Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -16,30 +17,72 @@ public class HttpBasedAgentContextService : IAgentContextService, IAgentContextI
...
+                FrontendClientName => await GetForFrontendClient(),
+                TravelGateConnectorClientName => await GetForApiClient(),
+                _ => default
+            };
+
+
+            async Task<AgentContext> GetForApiClient()
+            {
+                var name = GetHeaderValue("x-api-client-name");
+                var password = GetHeaderValue("x-api-client-password");
```
kirillta, line 56: Headers are case insensitive, so you have to normalize the value first. 
Also header canonical form is `X-Api-Client-Password`

kirillta (approved):

Api/Services/Agents/HttpBasedAgentContextService.cs
mingazhev: Changed to canonical form. Will discuss normalization later

PR #447, opened 2021-01-28
Added endpoint and new service for admins

AlexeyShok:
Added endpoint that returns agents list by agencyId


Api/AdministratorServices/AgentService.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public class AgentService : IAgentService
+    {
+        public AgentService(EdoContext context)
+        {
+            _context = context;
+        }
+
+
+        public async Task<Result<List<SlimAgentInfo>>> GetAgents(int agencyId)
```
mingazhev, line 21: Result is always success there

mingazhev (approved):

PR #449, opened 2021-01-29
Supporting multiple htIds in search

mingazhev:
Travelgate channel must support multiple destinations search, added this option in connection with mapper


Api/Models/Availabilities/AvailabilityRequest.cs
```diff
@@ -14,8 +14,8 @@ namespace HappyTravel.Edo.Api.Models.Availabilities
         [JsonConstructor]
         public AvailabilityRequest(string nationality, string residency, DateTime checkInDate, DateTime checkOutDate,
```
hasmikharutyunyan, line 15: is this model for our frontend request?

Api/Services/Accommodations/Availability/Mapping/AvailabilitySearchAreaService.cs
```diff
@@ -0,0 +1,63 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CSharpFunctionalExtensions;
+using HappyTravel.Edo.Api.Models.Availabilities.Mapping;
+using HappyTravel.Edo.Api.Models.Locations;
+using HappyTravel.Edo.Common.Enums;
+
+namespace HappyTravel.Edo.Api.Services.Accommodations.Availability.Mapping
+{
+    public class AvailabilitySearchAreaService : IAccommodationMappingService
```
hasmikharutyunyan, line 10: Why changed name ?

Api/Models/Availabilities/AvailabilityRequest.cs
mingazhev: Yes

Api/Services/Accommodations/Availability/Mapping/AvailabilitySearchAreaService.cs
mingazhev: In previous version it returned a single LocationDescriptor, info about the location.
In this version I think the data is semantically different, it contains multiple locations.
I've tried to return multiple LocationDescriptors but it was hard to work with them

hasmikharutyunyan: so, maybe needed to change IAccommodationMappingService name also?

mingazhev: Sure, missed that, now fixed

hasmikharutyunyan (approved):

oleg1std (approved):

oleg1std (approved):

PR #458, opened 2021-02-08
Bugfix/nijo 1236 payment processing service tries cancel booking after check in

AlexeyShok:
Added check for the CheckIn date when canceling bookings by processing.


mingazhev (commented):
We should not hide such bookings (after check in).
We must not allow this situation to occur - when booking is trying to cancel after check in.
Cancellation must occur before

mingazhev:
Closing, will rethink later

PR #459, opened 2021-02-09
Bugfix/nijo 1217 person information is not updated

AlexeyShok:
After updating the agent data, the agent's internal context data is updated

### DIFF
--- a/Api/Services/Agents/HttpBasedAgentContextService.cs
+++ b/Api/Services/Agents/HttpBasedAgentContextService.cs
@@ -40,14 +40,8 @@ namespace HappyTravel.Edo.Api.Services.Agents
         }
 
 
-        public async Task<Result> RefreshAgentContext()
-        {
+        public async Task RefreshAgentContext() =>
             _currentAgentContext = await GetAgentContext();
-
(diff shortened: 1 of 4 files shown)


kirillta (approved):

Api/Services/Agents/IAgentContextInternal.cs
```diff
@@ -7,5 +7,7 @@ namespace HappyTravel.Edo.Api.Services.Agents
     public interface IAgentContextInternal
     {
         ValueTask<Result<AgentContext>> GetAgentInfo();
+        
+        Task<Result> UpdateAgentContext();
```
mingazhev, line 11: looks more like "Refresh"

Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -40,6 +40,16 @@ public async ValueTask<Result<AgentContext>> GetAgentInfo()
         }
 
 
+        public async Task<Result> UpdateAgentContext()
```
mingazhev, line 43: No need to return result there. If agent context exist already, it will be refreshed anyway.
Method can be shortened to one line

PR #462, opened 2021-02-10
Added check if we need to remove old key

AlexeyShok:
Now the key is removed from the cache and reloaded only when needed

### DIFF
--- a/Api/Services/Agents/HttpBasedAgentContextService.cs
+++ b/Api/Services/Agents/HttpBasedAgentContextService.cs
@@ -40,21 +40,35 @@ namespace HappyTravel.Edo.Api.Services.Agents
         }
 
 
-        public async Task RefreshAgentContext() =>
-            _currentAgentContext = await GetAgentContext(isNeedReload: true);
+        public async Task RefreshAgentContext()
+        {
+            var identityClaim = _tokenInfoAccessor.GetIdentity();
+            var identityHash = identityClaim is not null
+                ? HashGenerator.ComputeSha256(identityClaim)
+                : string.Empty;
+
+            var key = _flow.BuildKey(nameof(HttpBasedAgentContextService), nameof(GetAgentInfo), identityHash);
+
+            await _flow.RemoveAsync(key);
+
+            _currentAgentContext = await _flow.GetOrSetAsync(
+                key: key,
+                getValueFunction: async () => await GetAgentInfoByIdentityHash(identityHash),
+                AgentContextCacheLifeTime);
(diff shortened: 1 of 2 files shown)


mingazhev (changes requested):
When using double flow you will have no way to reset caches on all EDO instances.

I suggest another way:
1. Change double flow to memory flow
2. Reset the cache manually (remove a key) when needed (in controller method that updates agent info), with calling a special method

Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -40,8 +40,22 @@ public async ValueTask<Result<AgentContext>> GetAgentInfo()
...
-        public async Task RefreshAgentContext() =>
-            _currentAgentContext = await GetAgentContext();
+        public async Task RefreshAgentContext()
+        {
+            var identityClaim = _tokenInfoAccessor.GetIdentity();
+            var identityHash = identityClaim is not null
+                ? HashGenerator.ComputeSha256(identityClaim)
+                : string.Empty;
+
+            var key = _flow.BuildKey(nameof(HttpBasedAgentContextService), nameof(GetAgentInfo), identityHash);
```
mingazhev, line 50: Key generation must be extracted to method

mingazhev, line 50: Is double flow changed to memory flow?

Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -40,8 +40,22 @@ public async ValueTask<Result<AgentContext>> GetAgentInfo()
...
+            var identityClaim = _tokenInfoAccessor.GetIdentity();
+            var identityHash = identityClaim is not null
+                ? HashGenerator.ComputeSha256(identityClaim)
+                : string.Empty;
+
+            var key = GetKey(identityHash);
+
+            _flow.Remove(key);
+
+            _currentAgentContext = await _flow.GetOrSetAsync(
```
mingazhev, line 54: Memory flow is synchronous, not need to use async overload

mingazhev (approved):

kirillta (approved):

Api/Services/Agents/HttpBasedAgentContextService.cs
AlexeyShok: Func in GetOrSetAsync is asynchronous, Func in GetOrSet is synchronous

PR #463, opened 2021-02-10
Added counterparty account add notification

Raibo:


Api/AdministratorServices/CounterpartyBillingNotificationService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                var defaultAgency = await _counterpartyService.GetDefaultAgency(counterpartyId);
+                return await _agentService.GetMasterAgent(defaultAgency.Id)
+                    .Map(master => master.Email);
+            }
+
+
+            Task<Result> SendNotification(string email)
+            {
+                var payload = new CounterpartyAccountAddedNotificationData
```
mingazhev, line 51: I think the whole operation should not fail if mail sending failed. Maybe just write error to the log?

Raibo: If we make possible that money addition succeed, but mail notification fails, then what result should be returned to controller? Success or failure?

Raibo: Done

hasmikharutyunyan (approved):

mingazhev (approved):

PR #465, opened 2021-02-11
fixed api client caching

DFSko:


kirillta (approved):

Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -76,11 +78,11 @@ async Task<AgentContext> GetForApiClient()
                 if (string.IsNullOrWhiteSpace(name) || string.IsNullOrWhiteSpace(password))
                     return default;
                 
-                var key = GetKey(name);
+                var key = GetKey($"name{HashGenerator.ComputeSha256(password)}");
```
mingazhev, line 81: why calculating hash two times?

mingazhev: Implementation is incorrect too

mingazhev (changes requested):

Api/Services/Agents/HttpBasedAgentContextService.cs
DFSko: Fixed

mingazhev (approved):

PR #466, opened 2021-02-11
NIJO-1267. Set Completed status if accommodation codes not found

oleg1std:
Set NotFound status for a supplier's response if the supplier's accommodation codes not found

### DIFF
--- a/Api/Services/Accommodations/Availability/Steps/WideAvailabilitySearch/WideAvailabilitySearchService.cs
+++ b/Api/Services/Accommodations/Availability/Steps/WideAvailabilitySearch/WideAvailabilitySearchService.cs
@@ -143,7 +143,7 @@ namespace HappyTravel.Edo.Api.Services.Accommodations.Availability.Steps.WideAva
                 accommodationCodes.TryGetValue(supplier, out var supplierCodeMappings);
                 if (supplierCodeMappings == null || !supplierCodeMappings.Any())
                 {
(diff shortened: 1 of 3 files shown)


Api/Services/Accommodations/Availability/Steps/WideAvailabilitySearch/WideAvailabilitySearchService.cs
```diff
@@ -132,20 +134,32 @@ IEnumerable<WideAvailabilityResult> CombineAvailabilities(IEnumerable<(Suppliers
...
+
+        private async Task StartSearch(Guid searchId, AvailabilityRequest request, AccommodationBookingSettings searchSettings,
             Location location, Dictionary<Suppliers, List<SupplierCodeMapping>> accommodationCodes, AgentContext agent, string languageCode)
         {
             foreach (var supplier in searchSettings.EnabledConnectors)
             {
                 accommodationCodes.TryGetValue(supplier, out var supplierCodeMappings);
+                if (supplierCodeMappings == null || !supplierCodeMappings.Any())
+                {
+                    await _availabilityStorage.SaveState(searchId, SupplierAvailabilitySearchState.NotFound(searchId), supplier);
```
mingazhev, line 146: I think we should return completed with 0 results there. NotFound looks like an invalid state and I'd prefer to remove it from enum

oleg1std: ok

kirillta (approved):

PR #482, opened 2021-02-19
Setting contract type during verification step

mingazhev:
Setting contract type as a mandatory step of verification

### DIFF
--- a/HappyTravel.Edo.Data/Agents/CounterpartyContractType.cs
+++ b/HappyTravel.Edo.Data/Agents/CounterpartyContractType.cs
@@ -1,0 +1,8 @@
+namespace HappyTravel.Edo.Data.Agents
+{
+    public enum CounterpartyContractType
+    {
+        CashPayments = 1,
+        CreditPayments = 2
+    }
+}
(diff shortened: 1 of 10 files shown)


HappyTravel.Edo.Data/Agents/CounterpartyContractType.cs
```diff
@@ -0,0 +1,8 @@
+namespace HappyTravel.Edo.Data.Agents
+{
+    public enum CounterpartyContractType
```
kirillta, line 3: Why not `PaymentType`?

mingazhev: I think of contract type as a set of predefined rules. Payment type can be confusing because credit customers may pay with a cash and etc., contract type has wider meaning.

Also from the admin panel backoffice staff will choose from one of the predefined contracts, not the payment type or set other contract parameters that we can implement later.

kirillta: Contract type could be confusing because of a booking contract as well

mingazhev: CounterpartyContractKind?

kirillta: Well, that's sounds different at least

kirillta (approved):

PR #483, opened 2021-02-19
Technical/nijo 1322 add agency id to agent markup2

Raibo:
This is second part of a partial pull request.
Added `agencyId` into all read/write places of `MarkupPolicies` table when dealing with agents markups.
Also added audit logging to admin service and refactored scope validation a bit.

### DIFF
--- a/Api/AdministratorServices/AgentService.cs
+++ b/Api/AdministratorServices/AgentService.cs
@@ -37,8 +37,8 @@ namespace HappyTravel.Edo.Api.AdministratorServices
                        Name = name,
                        Created = created,
                        IsActive = relation.IsActive,
-                       MarkupSettings = (agent.DisplayedMarkupFormula != null)
-                           ? agent.DisplayedMarkupFormula
+                       MarkupSettings = !string.IsNullOrWhiteSpace(relation.DisplayedMarkupFormula)
+                           ? relation.DisplayedMarkupFormula
                            : string.Empty
                    };
 
--- a/Api/Controllers/AgentControllers/AgentMarkupsController.cs
+++ b/Api/Controllers/AgentControllers/AgentMarkupsController.cs
@@ -111,7 +111,8 @@ namespace HappyTravel.Edo.Api.Controllers.AgentControllers
         [ProducesResponseType(typeof(List<MarkupInfo>), (int) HttpStatusCode.NoContent)]
         public async Task<IActionResult> GetPolicies(int agentId)
         {
-            return Ok(await _policyManager.Get(agentId));
+            var agencyId = (await _agentContext.GetAgent()).AgencyId;
+            return Ok(await _policyManager.Get(agentId, agencyId));
         }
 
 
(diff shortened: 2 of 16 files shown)


Api/AdministratorServices/AgentService.cs
```diff
@@ -37,8 +37,8 @@ public async Task<Result<List<SlimAgentInfo>>> GetAgents(int agencyId)
                        Name = name,
                        Created = created,
                        IsActive = relation.IsActive,
-                       MarkupSettings = (agent.DisplayedMarkupFormula != null)
-                           ? agent.DisplayedMarkupFormula
+                       MarkupSettings = !string.IsNullOrWhiteSpace(relation.DisplayedMarkupFormula)
```
mingazhev, line 40: Have you tested that this works?

Api/Controllers/AgentControllers/AgentMarkupsController.cs
```diff
@@ -111,7 +111,8 @@ public async Task<IActionResult> ModifyPolicy(int agentId, int policyId, [FromBo
         [ProducesResponseType(typeof(List<MarkupInfo>), (int) HttpStatusCode.NoContent)]
```
mingazhev, line 111: NoContent looks invalid

HappyTravel.Edo.Data/Migrations/20210219145917_MoveMarkupFormulaToRelation.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace HappyTravel.Edo.Data.Migrations
+{
+    public partial class MoveMarkupFormulaToRelation : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropColumn(
```
mingazhev, line 9: I think you should add a script that copies existing values. Or at least look at the production to be sure that we do not lose any data

Api/AdministratorServices/AgentService.cs
Raibo: `string.IsNullOrWhiteSpace`?
Numerous times.
This particular place in code - no, but `!string.IsNullOrWhiteSpace(s)` generally replaces `s != null` well.

Raibo: Takled a bit in discord - question was about convertion of this to SQL.
Answer - `!string.IsNullOrWhiteSpace` indeed converts to SQL and works fine.

HappyTravel.Edo.Data/Migrations/20210219145917_MoveMarkupFormulaToRelation.cs
Raibo: Done

Api/Controllers/AgentControllers/AgentMarkupsController.cs
Raibo: Fixed

mingazhev (approved):

PR #488, opened 2021-02-24
Cleanup and fix booking permissions

mingazhev:
The main change is that we always get booking from the database and then check its permissions.
This may be little longer to return that booking does not exist but looks more easy to understand and manage.

The next idea is to divide services to "internal" that can be used by anyone, and external, depending on the client (agent, admin etc.).
This allows less passing AgentContext through services and using it on the top level only. In this PR - all booking documents and sending that are available to use by an agent - extracted to a separate service and permissions are checked only there.
For background and other tasks - internal services are used.


Api/Services/Accommodations/Bookings/Management/BookingInfoService.cs
```diff
@@ -87,13 +87,13 @@ public async Task<Result<AccommodationBookingInfo>> GetAgentAccommodationBooking
 
         public async Task<Result<AccommodationBookingInfo>> GetAgentAccommodationBookingInfo(string referenceCode, AgentContext agentContext, string languageCode)
         {
-            var bookingDataResult = await Get(booking => agentContext.AgentId == booking.AgentId && booking.ReferenceCode == referenceCode);
+            var bookingDataResult = await _bookingRecordManager.Get(referenceCode)
+                .CheckPermissions(agentContext);
+
```
oleg1std, line 92: an extra line

oleg1std (approved):

AlexeyShok (approved):

Api/Services/Accommodations/Bookings/Management/BookingInfoService.cs
mingazhev: Fixed

kirillta (approved):

PR #2, opened 2020-02-25
Feature/nijo 405 update predictions

hasmikharutyunyan:

PR #489, opened 2021-02-24
Add childAgencies endpoint for agents and admins (#2)

Raibo:
https://github.com/happy-travel/agent-app-project/issues/2

### DIFF
--- a/Api/AdministratorServices/CounterpartyManagementService.cs
+++ b/Api/AdministratorServices/CounterpartyManagementService.cs
@@ -173,6 +173,12 @@ namespace HappyTravel.Edo.Api.AdministratorServices
                     .Tap(() => WriteAgencyActivationToAuditLog(agencyId, reason)));
 
 
+        public Task<List<AgencyInfo>> GetChildAgencies(int parentAgencyId)
+            => _context.Agencies.Where(a => a.ParentId == parentAgencyId)
+                .Select(a => new AgencyInfo(a.Name, a.Id))
+                .ToListAsync();
+
+
         private async Task<Result<Agency>> GetAgency(int agencyId)
         {
             var agency = await _context.Agencies.FirstOrDefaultAsync(ag => ag.Id == agencyId);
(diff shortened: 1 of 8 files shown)


Api/Controllers/AdministratorControllers/AgenciesController.cs
```diff
@@ -105,7 +109,19 @@ public async Task<IActionResult> GetAgents([FromRoute] int agencyId)
         }
 
 
+        /// <summary>
+        ///     Gets child agencies.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("{agencyId}/childAgencies")]
```
mingazhev, line 116: child-agencies

Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -173,6 +173,12 @@ public Task<Result> ActivateAgency(int agencyId, string reason)
                     .Tap(() => WriteAgencyActivationToAuditLog(agencyId, reason)));
 
 
+        public Task<List<AgencyInfo>> GetChildAgencies(int parentAgencyId)
```
mingazhev, line 176: Not sure that we need admin endpoints for this

Api/Services/Agents/ICounterpartyService.cs
```diff
@@ -19,5 +19,7 @@ public interface ICounterpartyService
         Task<Result<AgencyInfo>> GetAgency(int agencyId, AgentContext agent);
 
         Task<Agency> GetDefaultAgency(int counterpartyId);
+
+        Task<List<AgencyInfo>> GetChildAgencies(AgentContext agent);
```
mingazhev, line 23: placing this inside a counterparty service looks not good for me. please consider moving to another place

Api/AdministratorServices/CounterpartyManagementService.cs
Raibo: All the information should be available without DB, I think.

Api/Services/Agents/ICounterpartyService.cs
Raibo: Since methods like GetAgency are located here, this should too.
If we need to move this to a separate service, then we should move all related to agency as well, and this is refactoring.
Let's not combine new features with refactoring.

Api/Controllers/AdministratorControllers/AgenciesController.cs
Raibo: fixed

Api/Services/Agents/ICounterpartyService.cs
Raibo: Also, if we really need to refactor this, tell me or make a task, and I will move the methods. Similar situation with admin service.

AlexeyShok (approved):

PR #490, opened 2021-02-25
added ability to capture payments not only per day

DFSko:


mingazhev (commented):
I think we should merge this when all connectors will support time for cancellation policies

kirillta (approved):

DFSko:
not actual

PR #512, opened 2021-03-05
Proper way to act with failed booking

mingazhev:
Moved all booking changes logic to a single place + setting booking as invalid when it failed to book.

The main idea is to have a single place that is responsible for database booking updates: BookingRecordsUpdater.
The other responsibility is to act with connector: BookingManagementService.


Api/AdministratorServices/AdministratorBookingManagementService.cs
```diff
@@ -24,7 +31,7 @@ public Task<Result> Discard(int bookingId, Administrator administrator)
 
             
             Task<Result> ProcessDiscard(Booking booking) 
-                => _managementService.Discard(booking, administrator.ToUserInfo());
+                => _recordsUpdater.ChangeStatus(booking, BookingStatuses.Discarded, _dateTimeProvider.UtcNow(), administrator.ToUserInfo());
```
DFSko, line 34: Why we need to pass a whole object?

DFSko (approved):

PR #522, opened 2021-03-09
Technical/return available payment methods to frontend

mingazhev:
resolves https://github.com/happy-travel/agent-app-project/issues/30


mingazhev:
> Doesn't this PR requires additional tests?

Added unit tests

PR #528, opened 2021-03-12
Feature/aa 24 add writing booking change log

AlexeyShok:
Added logging booking statuses. I need to add the reason for the status change and logging other booking changes yet

### DIFF
--- a/Api/AdministratorServices/AdministratorBookingManagementService.cs
+++ b/Api/AdministratorServices/AdministratorBookingManagementService.cs
@@ -31,7 +31,7 @@ namespace HappyTravel.Edo.Api.AdministratorServices
 
             
             Task<Result> ProcessDiscard(Booking booking) 
-                => _recordsUpdater.ChangeStatus(booking, BookingStatuses.Discarded, _dateTimeProvider.UtcNow(), administrator.ToUserInfo(), new BookingChangeReason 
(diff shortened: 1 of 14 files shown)


Api/AdministratorServices/AdministratorBookingManagementService.cs
```diff
@@ -77,7 +85,11 @@ public async Task<Result> RejectManually(int bookingId, string reason, Administr
 
 
             Task<Result> RejectManually(Booking booking)
-                => _recordsUpdater.ChangeStatus(booking, BookingStatuses.Rejected, _dateTimeProvider.UtcNow(), admin.ToUserInfo());
+                => _recordsUpdater.ChangeStatus(booking, BookingStatuses.Rejected, _dateTimeProvider.UtcNow(), admin.ToUserInfo(), new BookingChangeReason 
+                { 
+                    ChangeSource = ChangeSources.Administrator,
```
mingazhev, line 90: Could we add a textual reason there to use with admin manual endpoints?

mingazhev (approved):

Api/AdministratorServices/AdministratorBookingManagementService.cs
AlexeyShok: Added

PR #535, opened 2021-03-16
added booking prices to analytics

DFSko:


Api/Models/Availabilities/Events/AccommodationBookingEvent.cs
```diff
@@ -18,6 +18,7 @@ namespace HappyTravel.Edo.Api.Models.Availabilities.Events
             SearchId = searchId;
             ResultId = resultId;
             RoomContractSetId = roomContractSetId;
+            Price = price;
```
mingazhev, line 21: Don't we need a currency there?

DFSko: I think elasticsearch will not to be able to convert currencies in calculations and all prices need to be stored in one currency

DFSko: Renamed

mingazhev (approved):

PR #536, opened 2021-03-16
implemented agencies production report

DFSko:


Api/Models/Reports/AgencyProduction.cs
```diff
@@ -0,0 +1,12 @@
+namespace HappyTravel.Edo.Api.Models.Reports
+{
+    public readonly struct AgencyProduction
```
konstyefremov, line 3: I think the word "production" isn't the best choice here. It usually relates to the production of material goods rather than services. Maybe "output" or "yield" will be a better choice.

Api/Models/Reports/AgencyProduction.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public string AgencyName { get; init; }
+        public int TotalBookings { get; init; }
```
konstyefremov, line 6: I think it's better to remove the word "Total". Otherwise the naming will be inconsistent. You should use the prefix "Total" for all totalled entities: Bookings, Revenue, Nights. Or not use it at all.

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+                .Map(GetRecords)
+                .Bind(Generate);
+
+
+            Result Validate()
+            {
+                if (fromDate == default || endDate == default)
+                    return Result.Failure<Stream>("Range dates required");
+            
+                if ((fromDate - fromDate).TotalDays > MaxRange)
```
konstyefremov, line 33: fromDate - fromDate = 0

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+                await _csvWriter.NextRecordAsync();
+                await _streamWriter.FlushAsync();
+            }
+            
+            stream.Seek(0, SeekOrigin.Begin);
+            return stream;
+        }
+        
+        
+        private const int MaxRange = 31;
```
konstyefremov, line 89: MaxRange might be more descriptive. Like MaxDaysInReport.

konstyefremov (changes requested):
if ((fromDate - fromDate).TotalDays > MaxRange) should be fixed

Api/Services/Reports/AgenciesProductionReportService.cs
DFSko: Fixed

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+                        Currency = grouped.Key.Currency.ToString(),
+                        Revenue = grouped.Sum(b => b.TotalPrice),
+                        Nights = grouped.Sum(b => (b.CheckOutDate - b.CheckInDate).Days),
+                        IsActive = grouped.Key.IsActive
+                    };
+            }
+        }
+        
+        
+        private async Task<Result<Stream>> Generate<TRow>(IEnumerable<TRow> records)
```
mingazhev, line 67: Isn't this a reusable logic to be used with another reports?

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            var stream = new MemoryStream();
```
AlexeyShok, line 69: Could it be better to use Using for streams?

DFSko: we cannot close the stream while the client downloading file. it should dispose after service disposing

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+                .Bind(Generate);
+
+
+            Result Validate()
+            {
+                if (fromDate == default || endDate == default)
+                    return Result.Failure<Stream>("Range dates required");
+            
+                if ((endDate - fromDate).TotalDays > MaxRange)
+                    return Result.Failure<Stream>("Permissible interval exceeded");
```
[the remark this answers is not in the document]
DFSko: Fixed

Api/Services/Reports/AgenciesProductionReportService.cs
```diff
@@ -0,0 +1,103 @@
...
+            
+            IQueryable<AgencyProduction> GetRecords()
+            {
+                var bookingsQuery = _context.Bookings
+                    .Where(b => b.Created >= fromDate && b.Created < endDate);
+                
+                return from agency in _context.Agencies
+                    join bookings in bookingsQuery on agency.Id equals bookings.AgencyId into joined
+                    from booking in joined.DefaultIfEmpty()
+                    group booking by new
```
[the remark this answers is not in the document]
DFSko: Yup, i'm testing, working fine

Api/Models/Reports/AgencyProduction.cs
```diff
@@ -0,0 +1,12 @@
...
+        public string Currency { get; init; }
+        public decimal Revenue { get; init; }
+        public int Nights { get; init; }
```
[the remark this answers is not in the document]
DFSko: Fixed

Api/Models/Reports/AgencyProduction.cs
DFSko: Renamed

Api/Services/Reports/AgenciesProductionReportService.cs
DFSko: This method should be reused in another report only if not needed in-memory mappings

Api/Models/Management/Enums/AdministratorPermissions.cs
```diff
@@ -20,5 +20,6 @@ public enum AdministratorPermissions
         AgentManagement = 4096,
         BalanceManualCorrection = 8192,
         DirectConnectivityReport = 16384,
```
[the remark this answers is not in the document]
DFSko: Fixed

Api/Models/Reports/AgencyProduction.cs
mingazhev: @kirillta says AgencyProduction is good enough

Api/Services/Reports/AgenciesProductionReportService.cs
mingazhev: Agree, I think MaxDaysInReport will be better

mingazhev (approved):
With a small comment about maxrange

konstyefremov (approved):

PR #555, opened 2021-03-23
Added a method to get a list of counterparty accounts

mingazhev:
Returns a list of available accounts for admin panel

https://github.com/happy-travel/edo/issues/549

### DIFF
--- a/Api/Models/Management/CounterpartyAccountInfo.cs
+++ b/Api/Models/Management/CounterpartyAccountInfo.cs
@@ -6,5 +6,6 @@ namespace HappyTravel.Edo.Api.Models.Management
     {
         public int Id { get; init; }
         public Currencies Currency { get; init; }
+        public decimal Balance { get; init; }
     }
 }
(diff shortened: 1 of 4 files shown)


hasmikharutyunyan (approved):

Api/Models/Management/CounterpartyAccountInfo.cs
```diff
@@ -0,0 +1,10 @@
+using HappyTravel.Money.Enums;
+
+namespace HappyTravel.Edo.Api.Models.Management
+{
+    public readonly struct CounterpartyAccountInfo
+    {
```
hasmikharutyunyan, line 6: Maybe `balance` also needed?

mingazhev: Sounds good

mingazhev: Fixed

PR #566, opened 2021-03-25
Moved fields from counterparty to agency, so far

Raibo:
https://github.com/happy-travel/agent-app-project/issues/51

no tests
no migrations

this is main PR, other PRs of the task will be merged to here.


Api/Services/Agents/AgencyValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+using FluentValidation;
+using HappyTravel.Edo.Api.Infrastructure;
+using HappyTravel.Edo.Api.Models.Agencies;
+using HappyTravel.Edo.Api.Models.Agents;
+
+namespace HappyTravel.Edo.Api.Services.Agents
+{
+    public static class AgencyValidator
+    {
+        public static Result Validate(in CounterpartyCreateRequest counterpartyInfo)
```
mingazhev, line 11: This method does not look like in right place, it is not bound to agency

Api/Models/Agents/CounterpartyCreateRequest.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System.ComponentModel.DataAnnotations;
+using HappyTravel.EdoContracts.General.Enums;
+using HappyTravel.Money.Enums;
+using Newtonsoft.Json;
+
+
+namespace HappyTravel.Edo.Api.Models.Agents
+{
+    public readonly struct CounterpartyCreateRequest
```
mingazhev, line 9: Is this still CounterpartyCreateRequest?

mingazhev (commented):
@kirillta will clarify with Thushara, what fields we actually can move from Counterparty and advise

HappyTravel.Edo.Data/Agents/Counterparty.cs
```diff
@@ -9,21 +9,12 @@ public class Counterparty
...
-        public Currencies PreferredCurrency { get; set; }
         public PaymentMethods PreferredPaymentMethod { get; set; }
-        public string Website { get; set; }
         public CounterpartyStates State { get; set; }
         public DateTime Created { get; set; }
         public string VerificationReason { get; set; }
         public DateTime? Verified { get; set; }
         public DateTime Updated { get; set; }
-        public string BillingEmail { get; set; }
         public string VatNumber { get; set; }
```
mingazhev, line 18: Please move the VatNumber to agency too. We've just got a feedback from the accounts team

Api/Services/Agents/AgencyValidator.cs
Raibo: I would not like to have this `Validate` methods in different places. This is a code duplication already, but at least here it is easily understandable, because they are next to each other.

If we, for example, move this validations into services they are used in, then this would be hard to guess, that there is identical validation of same fields in different type somewhere else.

And then we decide to change the validation and end up changing it in one place, but not in another.

Do you have any ideas to resolve this problem?

Api/Models/Agents/CounterpartyCreateRequest.cs
Raibo: Yes, why not?
This used solely for creation of a new counterparty.
It has agency fields because agency is always created along with a counterparty.

Api/Services/Agents/AgencyValidator.cs
Raibo: Changed format of CounterpartyCreateRequest, now this should not be a problem.

HappyTravel.Edo.Data/Agents/Counterparty.cs
Raibo: Ok

Raibo:
Previously `CounterpartyManagementService.GetAllCounterpartyAgencies(counterpartyId)` returned failure, if there was no counterparty with Id `counterpartyId`.
Now it just returns empty list. 

If this is not okay, say something.
@mingazhev @kirillta

Raibo:
There is an endpoint to update a Counterparty, but there is no endpoint to update an Agency.
This means that now there is no way to update moved fields.

If this is not okay, say something.
@mingazhev @kirillta

HappyTravel.Edo.UnitTests/Tests/AdministratorServices/CounterpartyManagementServiceTests/CounterpartyManagementTests.cs
```diff
@@ -44,28 +44,27 @@ public async Task Get_not_existed_counterparty_should_fail()
         [Fact]
         public async Task Get_counterparty_should_return_all_counterparties()
         {
-            var counterpartyList = await _counterpartyManagementService.Get("en");
+            var counterpartyList = await _counterpartyManagementService.Get();
 
-            Assert.True(counterpartyList.Count == 3);
+            Assert.True(counterpartyList.Count == 5);
```
Raibo, line 49: We now expect count to be 5 because previously Counterparties with `LanguageCode == null` were ignored, therefore only 3 of 5 total Counterparties made it to the list.

Now there is no `CountryCode` in Counterparty, so all 5 appears in the list.

mingazhev (approved):

HappyTravel.Edo.Data/Agents/Counterparty.cs
```diff
@@ -1,30 +1,19 @@
...
 using HappyTravel.EdoContracts.General.Enums;
-using HappyTravel.Money.Enums;
 
 namespace HappyTravel.Edo.Data.Agents
 {
     public class Counterparty
     {
         public int Id { get; set; }
         public string Name { get; set; }
-        public string Address { get; set; }
```
kirillta, line 12: As discussed, some fields should remain in this table

PR #52, opened 2020-04-24
NIJO-573. Fix saving BookingDetails on the cancellation process

oleg1std:

PR #590, opened 2021-04-01
Add endpoint to edit self agency fields

Raibo:
https://github.com/happy-travel/agent-app-project/issues/52

### DIFF
--- a/Api/Models/Agencies/EditAgencyInfo.cs
+++ b/Api/Models/Agencies/EditAgencyInfo.cs
@@ -1,0 +1,48 @@
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Models.Agencies
+{
+    public readonly struct EditAgencyInfo
+    {
+        [JsonConstructor]
+        public EditAgencyInfo(string address, string billingEmail, string fax, string phone, string postalCode, string website)
+        {
+            Address = address;
+            BillingEmail = billingEmail;
+            Fax = fax;
+            Phone = phone;
+            PostalCode = postalCode;
+            Website = website;
+        }
+
+        /// <summary>
+        ///     Agency address.
+        /// </summary>
+        public string Address { get; }
+
+        /// <summary>
+        ///     Phone number. Only digits, length between 3 and 30.
+        /// </summary>
+        public string Phone { get; }
+
+        /// <summary>
+        ///     Fax number. Only digits, length between 3 and 30.
+        /// </summary>
+        public string Fax { get; }
+
+        /// <summary>
+        ///     Postal code.
+        /// </summary>
+        public string PostalCode { get; }
+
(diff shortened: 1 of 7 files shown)


Api/Models/Agencies/EditAgencyInfo.cs
```diff
@@ -0,0 +1,48 @@
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Models.Agencies
+{
+    public readonly struct EditAgencyInfo
+    {
+        [JsonConstructor]
+        public EditAgencyInfo(string address, string billingEmail, string fax, string phone, string postalCode, string website)
```
mingazhev, line 8: This should be EditAgencyRequest or EditableAgencyInfo (as it is in another places)

Api/Services/Agents/AgencyService.cs
```diff
@@ -34,6 +40,43 @@ public Task<Result<SlimAgencyInfo>> Get(AgentContext agent, string languageCode
...
+            async Task UpdateAgencyRecord()
+            {
+                var agencyRecord = await _edoContext.Agencies.SingleAsync(a => a.Id == agent.AgencyId);
+
+                agencyRecord.Address = editAgencyInfo.Address ?? agencyRecord.Address;
+                agencyRecord.Phone = editAgencyInfo.Phone ?? agencyRecord.Phone;
+                agencyRecord.Fax = editAgencyInfo.Fax ?? agencyRecord.Fax;
+                agencyRecord.PostalCode = editAgencyInfo.PostalCode ?? agencyRecord.PostalCode;
+                agencyRecord.Website = editAgencyInfo.Website ?? agencyRecord.Website;
+                agencyRecord.BillingEmail = editAgencyInfo.BillingEmail ?? agencyRecord.BillingEmail;
```
mingazhev, line 64: This is not correct since you use a PUT method. PUT method updates the whole entity.
And I don't think that it is useful to update only parts there

mingazhev, line 64: This is not correct since you use a PUT method. PUT method updates the whole entity.
And I don't think that it is useful to update only parts there

Api/Services/Agents/AgencyValidator.cs
```diff
@@ -18,5 +18,16 @@ public static Result Validate(in AgencyInfo agencyInfo)
                 v.RuleFor(c => c.Fax).Matches(@"^[0-9]{3,30}$").When(i => !string.IsNullOrWhiteSpace(i.Fax));
             }, agencyInfo);
         }
+
+
+        public static Result Validate(in EditAgencyRequest agencyRequest)
```
mingazhev, line 23: Think this method will be better to place nearer to usage

mingazhev (approved):

PR #592, opened 2021-04-02
Bugfix/edo 537 agent id is not logged in booking status history table part2

AlexeyShok:
Added logging status for create booking. Added UserInfo parameter for ProcessResponse method.


kirillta (approved):

mingazhev (approved):
Let's merge this and I'll inspect this additionally on my own

PR #594, opened 2021-04-05
Replace tags with direct connectivity flag

mingazhev:


Api/Services/Accommodations/Availability/Steps/BookingEvaluation/BookingEvaluationService.cs
```diff
@@ -143,11 +143,11 @@ Task SaveToCache(DataWithMarkup<EdoContracts.Accommodations.RoomContractSetAvail
                     ? result.Supplier
                     : (Suppliers?) null;
 
-                var tags = settings.AreTagsVisible
-                    ? availabilityData.Value.RoomContractSet.Tags
-                    : new List<string>(0);
+                bool? isDirectContract = settings.IsDirectContractFlagVisible
```
kirillta, line 146: I don't think the field should be nullable, because of such logic. As for me, the value should be set to false if flags are invisible

Api/Services/Accommodations/Availability/Steps/RoomSelection/RoomSelectionService.cs
```diff
@@ -150,11 +150,11 @@ IEnumerable<RoomContractSet> MapToRoomContractSets(SupplierData<AccommodationAva
                             ? accommodationAvailability.Source
                             : (Suppliers?) null;
 
-                        var tags = searchSettings.AreTagsVisible
-                            ? rs.Tags
-                            : new List<string>();
+                        bool? isDirectContractFlag = searchSettings.IsDirectContractFlagVisible
```
kirillta, line 153: Same here

Api/Services/Accommodations/Availability/Steps/BookingEvaluation/BookingEvaluationService.cs
kirillta: The uncertainness of the always false statement is understandable, but null is not better. In fact, the field not represents the actual contract state, but only its appearance

mingazhev: changed to non-nullable

kirillta (approved):

PR #599, opened 2021-04-05
Getting initiator type from user info

mingazhev:
To not pass it to parameters every time

### DIFF
--- a/HappyTravel.Edo.Common/Enums/UserTypes.cs
+++ b/HappyTravel.Edo.Common/Enums/UserTypes.cs
@@ -10,6 +10,7 @@ namespace HappyTravel.Edo.Common.Enums
         Admin = 1,
         Agent = 2,
         ServiceAccount = 3,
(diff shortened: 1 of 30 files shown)


HappyTravel.Edo.Common/Enums/UserTypes.cs
```diff
@@ -10,6 +10,7 @@ public enum UserTypes
         Admin = 1,
         Agent = 2,
         ServiceAccount = 3,
-        InternalServiceAccount = 4
+        InternalServiceAccount = 4,
+        Supplier = 5
```
AlexeyShok, line 14: It seems to me that adding a supplier to user types is incorrect.

hasmikharutyunyan: maybe change `UserTypes` to another name?

DFSko (approved):

PR #604, opened 2021-04-07
added second booking price with commission

DFSko:
added second booking price with commission

resolves happy-travel/agent-app-project#114


Api/Services/Accommodations/Bookings/BookingExecution/BookingRegistrationService.cs
```diff
@@ -168,6 +168,7 @@ void AddServiceDetails()
             {
                 var rate = availabilityInfo.RoomContractSet.Rate;
                 booking.TotalPrice = rate.FinalPrice.Amount;
+                booking.CreditCardPaymentPrice = rate.CreditCardPrice.Amount;
```
kirillta, line 171: Ideally both prices should be in `MoneyAmount`

kirillta (approved):

Api/Services/Accommodations/Bookings/BookingExecution/BookingRegistrationService.cs
DFSko: We can't store MoneyAmount in database

Api/Models/Accommodations/Rate.cs
```diff
@@ -45,6 +46,12 @@ namespace HappyTravel.Edo.Api.Models.Accommodations
         ///     The final and total net price of a service. This is <b>the actual</b> value of a price.
         /// </summary>
         public MoneyAmount FinalPrice { get; }
+        
+        
+        /// <summary>
```
mingazhev, line 51: Don't think that we need a separate property for this

Api/Models/Accommodations/Rate.cs
```diff
@@ -10,13 +10,14 @@ namespace HappyTravel.Edo.Api.Models.Accommodations
     public readonly struct Rate
     {
         [JsonConstructor]
-        public Rate(in MoneyAmount finalPrice, in MoneyAmount gross, List<Discount> discounts,
+        public Rate(in MoneyAmount finalPrice, in MoneyAmount creditCardPrice, in MoneyAmount gross, List<Discount> discounts,
```
mingazhev, line 13: I think it will be better if final price will be a dictionary Dictionary<PaymentMethods, MoneyAmount>

mingazhev (approved):
Looks fine for me now, but we need to wait frontend to make changes before merge

kirillta (approved):

DFSko:
not actual

PR #623, opened 2021-04-14
Technical/change some admin api places

hasmikharutyunyan:
Merged another PR with admin api changes here


Api/Controllers/AdministratorControllers/AgenciesController.cs
```diff
@@ -97,8 +118,8 @@ public async Task<IActionResult> SetSystemSettings([FromBody] AgencyAccommodatio
         /// <param name="agencyId">Agency Id</param>
         /// <returns>List of agents</returns>
         [HttpGet("{agencyId}/agents")]
-        [ProducesResponseType(typeof(List<SlimAgentInfo>), (int)HttpStatusCode.OK)]
-        [ProducesResponseType(typeof(ProblemDetails), (int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType(typeof(List<SlimAgentInfo>), (int) HttpStatusCode.OK)]
```
DFSko, line 121: we can use Microsoft.AspNetCore.Http.StatusCodes.Status200OK without casting type

Api/Controllers/AdministratorControllers/AgentsController.cs
```diff
@@ -20,14 +22,36 @@ namespace HappyTravel.Edo.Api.Controllers.AdministratorControllers
...
+
+        /// <summary>
+        ///     Gets a list of all bookings made by the agent
+        /// </summary>
+        /// <param name="agentId">Agent Id</param>
+        /// <returns>List of bookings</returns>
+        [HttpGet("agents/{agentId}/accommodations/bookings")]
+        [ProducesResponseType(typeof(List<Booking>), (int) HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
+        [AdministratorPermissions(AdministratorPermissions.AgentManagement)]
```
DFSko, line 44: BookingManagement?

Api/Controllers/AdministratorControllers/InvitationsController.cs
```diff
@@ -37,7 +37,7 @@ public class ManagementController : BaseController
         /// </summary>
         /// <param name="invitationInfo">Administrator invitation info.</param>
         /// <returns></returns>
-        [HttpPost("invite")]
+        [HttpPost]
```
AlexeyShok, line 40: Maybe [HttpPost("send")] ?

hasmikharutyunyan: I think by POST request in InvitationsController is understandable that  request is for invite.

Api/Controllers/AdministratorControllers/AgenciesController.cs
```diff
@@ -23,11 +24,31 @@ public class AgenciesController : BaseController
...
+        /// <summary>
+        ///     Gets a list of all bookings made by the agency
+        /// </summary>
+        /// <param name="agencyId">Agency Id</param>
+        /// <returns>List of bookings</returns>
+        [HttpGet("{agencyId}/accommodations/bookings")]
+        [ProducesResponseType(typeof(List<Booking>), (int) HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
+        [AdministratorPermissions(AdministratorPermissions.AgentManagement)]
+        public async Task<IActionResult> GetAgencyBookings([FromRoute] int agencyId)
```
mingazhev, line 45: Don't think that we need to store booking methods in separate controllers by their owners.
In such case it will later be very large amount of methods related to agency data, all in one controller.
Agency bookings, invitations, maybe some logs and etc. Logically these resources belong to agency, but I don't think we should reflect such things in a controller.
I'll think again later

Api/Controllers/AdministratorControllers/AgenciesController.cs
hasmikharutyunyan: I have done in some places, but I think needed another pull request for that to change everywhere.

AlexeyShok (approved):

PR #627, opened 2021-04-14
Add attribute to serialize enums by name to registration mail info

Raibo:


Api/Models/Mailing/RegistrationDataForAdmin.cs
```diff
@@ -1,26 +1,41 @@
...
 
 
         public struct CounterpartyRegistrationMailData
         {
             public string Name { get; set; }
+
             public string Address { get; set; }
+
+            [JsonConverter(typeof(StringEnumConverter))]
             public PaymentTypes PreferredPaymentMethod { get; set; }
```
mingazhev, line 24: if this is mail model you should send a string that is got from applying enumconverter (it uses enum's description)

Raibo: Ok. I was just confused, because in history of the file there is nothing like that, and yet it worked somehow.

mingazhev (approved):

PR #632, opened 2021-04-17
fixed validation error on modify policy

DFSko:


mingazhev (changes requested):
Checking order is still needed even when updating policy

mingazhev (commented):
Need to change validation method, not passing a flag to ignore validation outside

mingazhev (approved):

PR #633, opened 2021-04-18
Added discounts applicable to markups

mingazhev:
In this variant Discount is a thing that can be applied to a certain markup and markup id is needed to create a discount.
This is simple enough to implement and understand and automatically supports right markup bonuses calculation:
Every discount is applied to a given markup and all the bonuses will be calculated with the resulting value (discounts are integrated to markup bonuses).

From the other hand, setting markup not only by agency but by a markup too maybe confusing and not fully understandable.
"Why I cannot just select an agency and give it discount? Why I need to select a policy to apply markup to?"

### DIFF
--- a/Api/Services/Markups/DiscountFunctionService.cs
+++ b/Api/Services/Markups/DiscountFunctionService.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using FloxDc.CacheFlow;
+using FloxDc.CacheFlow.Extensions;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Api.Services.PriceProcessing;
+using HappyTravel.Edo.Common.Enums.Markup;
+using HappyTravel.Edo.Data;
+using HappyTravel.Edo.Data.Markup;
+using HappyTravel.Money.Models;
+using Microsoft.EntityFrameworkCore;
+
+namespace HappyTravel.Edo.Api.Services.Markups
+{
+    public class DiscountFunctionService : IDiscountFunctionService
+    {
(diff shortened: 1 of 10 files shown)


HappyTravel.Edo.Data/Markup/Discount.cs
```diff
@@ -0,0 +1,11 @@
+namespace HappyTravel.Edo.Data.Markup
+{
+    public class Discount
+    {
+        public int Id { get; set; }
+        public int AgencyId { get; set; }
```
kirillta, line 6: Not clear, what `AgencyId` means here

Api/Services/Markups/DiscountFunctionService.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        
+        
+        public async ValueTask<PriceProcessFunction> Get(MarkupPolicy policy, AgentContext agent)
+        {
+            // Discounts are only supported for global markups for now
+            if (policy.ScopeType != MarkupPolicyScopeType.Global)
+                return (price => new ValueTask<MoneyAmount>(price));
+
+            var discountsKey = GetKey(policy, agent);
```
kirillta, line 32: discountKey

Api/Services/Markups/DiscountFunctionService.cs
```diff
@@ -0,0 +1,71 @@
...
+            var applicableDiscounts = await _flow.GetOrSetAsync(discountsKey, 
+                GetAgentDiscounts,
```
kirillta, line 34: Line-breaking formatting is too explicit. Could be on a line

kirillta:
Here I don't happy with how a discount becomes a part of a markup

mingazhev:
Closing PR, found a better approach

mingazhev:
This one is in master, some kind

PR #634, opened 2021-04-18
Independent discounts

mingazhev:
In this variant discounts are implemented similar to markup service and are applied independently.
More code there and more code to be written, but from the client side looks more clear to specify a discount for a single agency and do not think about other sides.

There it need to be stored some discount application log that should be later used to materialize markup bonuses: applied discounts should be substracted from markup bonuses during materialization.
This is not first class task since we do not materialize global and counterparty markups for now but this thing needs to be remembered.


Api/Services/Markups/IDiscountService.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public interface IDiscountService
+    {
+        Task<TDetails> ApplyDiscounts<TDetails>(AgentContext agent, TDetails details,
+            Func<TDetails, PriceProcessFunction, ValueTask<TDetails>> priceProcessFunc,
+            Action<DiscountApplicationResult<TDetails>> logAction = null);
+    }
+    
+
+    public readonly struct DiscountApplicationResult<TDetails>
```
kirillta, line 17: A struct and an interface in a same file

HappyTravel.Edo.Data/Markup/Discount.cs
```diff
@@ -0,0 +1,10 @@
+namespace HappyTravel.Edo.Data.Markup
+{
+    public class Discount
+    {
+        public int Id { get; set; }
+        public int AgencyId { get; set; }
```
kirillta, line 6: Not clear, what `AgencyId` means here

Api/Services/Markups/DiscountService.cs
```diff
@@ -0,0 +1,87 @@
...
+            _context = context;
+            _flow = flow;
+        }
+        
+        
+        public async Task<TDetails> ApplyDiscounts<TDetails>(AgentContext agent, TDetails details,
+            Func<TDetails, PriceProcessFunction, ValueTask<TDetails>> priceProcessFunc,
+            Action<DiscountApplicationResult<TDetails>> logAction)
+        {
+            var discountsKey = GetKey(agent);
```
kirillta, line 30: discountKey

Api/Services/Markups/DiscountService.cs
```diff
@@ -0,0 +1,87 @@
...
+            var applicableDiscounts = await _flow.GetOrSetAsync(discountsKey,
```
kirillta, line 31: Could be on a line

kirillta (approved):

kirillta:
This one is better, but we apply a discount after all markups. That's complex to calculate

Api/Services/Markups/DiscountService.cs
```diff
@@ -0,0 +1,87 @@
...
+        public async Task<TDetails> ApplyDiscounts<TDetails>(AgentContext agent, TDetails details,
+            Func<TDetails, PriceProcessFunction, ValueTask<TDetails>> priceProcessFunc,
+            Action<DiscountApplicationResult<TDetails>> logAction)
+        {
+            var discountsKey = GetKey(agent);
+            var applicableDiscounts = await _flow.GetOrSetAsync(discountsKey, 
+                GetAgentDiscounts,
+                DiscountCacheLifeTime);
+
+            var currentData = details;
```
kirillta, line 35: Usage of the word `data` is confuses here

kirillta (approved):

mingazhev (commented):
> 
> 
> This one is better, but we apply a discount after all markups. That's complex to calculate.
> Or it isn't. Depends on what `TDetails` are

TDetails is a simple struct with availability result, RoomContractSet, AccommodationAvailability and etc.
It's hard to calculate when we need a strict final markup (which is the case now) but is easy to understand that we get the final price and apply a discount to it.
For current situation we know that global markup is 6% (1.06).
To achieve 1.03 we should calculate a discount percent as `(1-1.03/1.06) * 100 = 2.83%`

I think this is the nature of discount, to apply in such way. If we need to apply strictly 3% and nothing else, there would be some exceptions from applying markup or some kind of "overriding markup". We've discussed such ways and decided not to use them

PR #636, opened 2021-04-19
Added discount management

mingazhev:


Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -0,0 +1,97 @@
...
+                .Check(_ => Validate(discountInfo))
+                .Map(d => Update(d, discount =>
+                {
+                    discount.DiscountPercent = discountInfo.DiscountPercent;
+                    discount.Description = discountInfo.Description;
+                }));
+        }
+
+
+        private async Task<Result<Discount>> Get(int agencyId, int discountId)
```
DFSko, line 74: why is id needed here?

mingazhev: not needed if being strict, but for meaningful routes including an agency

Raibo: as an option, we could include agencyId into the route, but not take as a parameter further, leavling it unused in the controller

mingazhev: I'm trying to avoid unused vars

mingazhev:
Another approach chosen

PR #644, opened 2021-04-21
Feature/aa 148 save additional accommodation info in booking record

AlexeyShok:
Added AccommodationInfo to Booking model. Added property Photo to AccommodationInfo model. These photo are saved in the BookingAvailabilityInfo at stage booking evaluation and then added to the booking when it is created.


HappyTravel.Edo.Data/Bookings/Booking.cs
```diff
@@ -45,6 +45,7 @@ public class Booking : IEntity
         
         public string AccommodationId { get; set; }
         public string AccommodationName { get; set; }
+        public AccommodationInfo AccommodationInfo { get; set; }
```
kirillta, line 48: Why we have an ID, a name, info, and a location in four difference objects? All of them supposed to be parts of `AccommodationInfo`

mingazhev: Ok, let's have all this in one place. But in a different PR because migration can be complex

mingazhev (commented):
Could not find much differences in 2 PRs, are they related?

HappyTravel.Edo.Data/Bookings/Booking.cs
AlexeyShok: Ok

mingazhev (approved):

PR #645, opened 2021-04-21
Feature/aa 149 return accommodation photo url in rich booking model

AlexeyShok:
Added AccommodationInfo with Photo to AccommodationDetails model. Based on feature/AA-148_SaveAdditionalAccommodationInfoInBookingRecord branch.


kirillta (commented):
`AccommodationInfo` should contain all related info in it

Api/Models/Accommodations/AccommodationInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using HappyTravel.EdoContracts.Accommodations.Internals;
+using Newtonsoft.Json;
+using System;
+
+namespace HappyTravel.Edo.Api.Models.Accommodations
+{
+    public readonly struct AccommodationInfo
+    {
+        [JsonConstructor]
+        public AccommodationInfo(ImageInfo photo)
```
mingazhev, line 10: Please do not use contracts (instead of enums) in frontend models. This causes lots of problems every time

mingazhev: By my experience it is fine to make a copy of a model to use inside EDO. At list typically it should be a class to work fine with EF

mingazhev (changes requested):

Api/Models/Accommodations/AccommodationInfo.cs
AlexeyShok: This model is not transferred to the front.

mingazhev (approved):

PR #647, opened 2021-04-22
Fixed migration and ISendingSettings serialization

AlexeyShok:
Fixed migration and ISendingSettings serialization


AlexeyShok:
Uncorrected EdoContextSnapshot

PR #651, opened 2021-04-26
Hotfix/aa 149 fix swagger exception

AlexeyShok:
Fix swagger exception

### DIFF
--- a/Api/Startup.cs
+++ b/Api/Startup.cs
@@ -133,6 +133,8 @@ namespace HappyTravel.Edo.Api
                         Array.Empty<string>()
                     }
                 });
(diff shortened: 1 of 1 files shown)


kirillta (approved):

Api/Startup.cs
```diff
@@ -133,6 +133,8 @@ public void ConfigureServices(IServiceCollection services)
                         Array.Empty<string>()
                     }
                 });
+                // Use fully qualified object names
```
mingazhev, line 136: Comment is not needed

mingazhev (approved):

PR #664, opened 2021-05-05
Get user registration email from identity

Raibo:
https://github.com/happy-travel/agent-app-project/issues/72

### DIFF
--- a/Api/Controllers/AdministratorControllers/InvitationsController.cs
+++ b/Api/Controllers/AdministratorControllers/InvitationsController.cs
@@ -1,13 +1,17 @@
+using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
 using HappyTravel.Edo.Api.AdministratorServices.Invitations;
 using HappyTravel.Edo.Api.Filters.Authorization.AdministratorFilters;
 using HappyTravel.Edo.Api.Infrastructure;
+using HappyTravel.Edo.Api.Infrastructure.Constants;
 using HappyTravel.Edo.Api.Infrastructure.Invitations;
 using HappyTravel.Edo.Api.Models.Management.Enums;
 using HappyTravel.Edo.Api.Models.Users;
 using HappyTravel.Edo.Api.Services.Management;
+using IdentityModel.Client;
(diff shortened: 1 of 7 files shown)


Api/Controllers/AgentControllers/AgenciesController.cs
```diff
@@ -204,10 +209,15 @@ public async Task<IActionResult> AcceptChildAgencyInvite([FromBody] RegisterInvi
             if (string.IsNullOrWhiteSpace(identity))
                 return BadRequest(ProblemDetailsBuilder.Build("User sub claim is required"));
 
+            var email = await GetUserEmail();
+            if (string.IsNullOrWhiteSpace(email))
+                return BadRequest(ProblemDetailsBuilder.Build("E-mail claim is required"));
```
kirillta, line 214: It's good to add _why_ it's required

kirillta: And above as well

Api/Controllers/AgentControllers/AgenciesController.cs
```diff
@@ -256,12 +266,28 @@ public async Task<IActionResult> ActivateAgency(int agencyId)
         }
 
 
+        private async Task<string> GetUserEmail()
+        {
+            // TODO: Move this logic to separate service
```
kirillta, line 271: I think this _TODO_ is totally doable. Why not to move this code now?

Api/Controllers/AdministratorControllers/InvitationsController.cs
```diff
@@ -92,12 +102,28 @@ public async Task<IActionResult> DisableInvitation(string codeHash)
 
             return Ok();
         }
-        
-        
+
+
+        private async Task<string> GetUserEmail()
```
kirillta, line 107: And here. Obviously no need to copy these lines three times

kirillta (changes requested):

Api/Controllers/AgentControllers/AgenciesController.cs
Raibo: I knew you'll get me on this one. Fixed.

kirillta (approved):

PR #670, opened 2021-05-07
added supplier reference code to invoice and voucher emails data

DFSko:


kirillta (approved):
This part is done, but not all suppliers support these codes. The field must be hided in an email itself as well

mingazhev (approved):

PR #672, opened 2021-05-08
fixed sentry EDO-API-A1 issue

DFSko:
Link to error: https://sentry1.happytravel.com/organizations/happytravel/issues/2294/?project=2&referrer=AssignedActivityEmail


mingazhev (commented):
1. please add a link to error
2. looks like it should be fixed in another place
3. we must know about such errors, maybe from sentry
4. not sure that reason phrase will contain any needed info

mingazhev (approved):

PR #683, opened 2021-05-21
Feature/aa 221 add notifications via notification center for counterparty

AlexeyShok:
Added notifications via Notification Center for customer invitation, regular customer successful registration and child agency successful registration

### DIFF
--- a/Api/NotificationCenter/Services/NotificationService.cs
+++ b/Api/NotificationCenter/Services/NotificationService.cs
@@ -105,6 +105,12 @@ namespace HappyTravel.Edo.Api.NotificationCenter.Services
         }
 
 
+        public async Task<Result> Send(DataWithCompanyInfo messageData, NotificationTypes notificationType, string email, string templateId)
+        {
+            return await Send(new SlimAgentContext(agentId: 0, agencyId: 0), messageData, notificationType, new List<string> { email }, templateId);
+        }
+
+
         private static Dictionary<ProtocolTypes, object> BuildSettings(SlimNotificationOptions notificationOptions, List<string> emails, string templateId)
         {
             var sendingSettings = new Dictionary<ProtocolTypes, object>();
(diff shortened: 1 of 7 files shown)


hasmikharutyunyan (approved):

Api/NotificationCenter/Services/NotificationService.cs
```diff
@@ -105,6 +105,12 @@ public async Task<Result> Send(SlimAgentContext agent, DataWithCompanyInfo messa
         }
 
 
+        public async Task<Result> Send(DataWithCompanyInfo messageData, NotificationTypes notificationType, string email, string templateId)
+        {
+            return await Send(new SlimAgentContext(agentId: 0, agencyId: 0), messageData, notificationType, new List<string> { email }, templateId);
```
hasmikharutyunyan, line 110: Why here agentId and agencyId are equals to 0 ?

Api/NotificationCenter/Services/NotificationService.cs
```diff
@@ -105,6 +105,12 @@ public async Task<Result> Send(SlimAgentContext agent, DataWithCompanyInfo messa
         }
 
 
+        public async Task<Result> Send(DataWithCompanyInfo messageData, NotificationTypes notificationType, string email, string templateId)
```
kirillta, line 108: A `DataWithCompanyInfo` name is a bit awkward

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: At the moment of sending the invitation, we do not know the agent ID. In the case of creating a subsidiary agency, we also do not yet know the Id of the subsidiary agency.

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: This name was invented a long time ago. This is the base class that adds information about the HappyTravel company to the email. I do not know how to call it otherwise. Offer options.

kirillta (approved):

oleg1std (approved):

PR #716, opened 2021-06-02
Added endpoint for update admin notification options

AlexeyShok:
Added endpoint for update admin notification options

### DIFF
--- a/Api/NotificationCenter/Services/NotificationOptionsService.cs
+++ b/Api/NotificationCenter/Services/NotificationOptionsService.cs
@@ -85,7 +85,7 @@ namespace HappyTravel.Edo.Api.NotificationCenter.Services
                 if (defaultOptions.IsFailure)
                     return Result.Failure<SlimNotificationOptions>(defaultOptions.Error);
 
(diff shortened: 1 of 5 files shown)


hasmikharutyunyan (approved):

Api/NotificationCenter/Services/NotificationOptionsService.cs
```diff
@@ -85,7 +85,7 @@ Result<SlimNotificationOptions> Validate()
                 if (defaultOptions.IsFailure)
                     return Result.Failure<SlimNotificationOptions>(defaultOptions.Error);
 
-                if (defaultOptions.Value.IsMandatory && options.EnabledProtocols != default)
+                if (defaultOptions.Value.IsMandatory && options.EnabledProtocols != defaultOptions.Value.EnabledProtocols)
```
kirillta, line 88: Can we return evaluated options without that value call?

kirillta (approved):

Api/NotificationCenter/Services/NotificationOptionsService.cs
AlexeyShok: No

PR #727, opened 2021-06-09
Technical/aa 201 verify routes for restfulness v1

hasmikharutyunyan:
Done changes which affect only our services. 
@mingazhev, in matsumato and payment processing we will change also, after that we will merge this pull request.


mingazhev (approved):
It's ok and now need to sync clients:
Matsumoto and payment processing

Api/Controllers/AgentControllers/PaymentsController.cs
```diff
@@ -59,7 +55,7 @@ public IActionResult GetPaymentTypes()
         ///     Pays by payfort token
         /// </summary>
         /// <param name="request">Payment request</param>
-        [HttpPost("bookings/card/new")]
+        [HttpPost("accommodations/bookings/pay-by-new-card")]
```
kirillta, line 58: Probably `accommodations/bookings/cards/new/pay`?

Api/Controllers/AgentControllers/InternalBookingsController.cs
```diff
@@ -140,11 +135,11 @@ public async Task<IActionResult> Charge(List<int> bookingIds)
         /// </summary>
         /// <param name="date">Deadline date</param>
         /// <returns>Result message</returns>
-        [HttpGet("notify/deadline-approach/{date}")]
+        [HttpGet("to-notify/deadline-approach")]
```
kirillta, line 138: Why `to`?

kirillta (changes requested):

Api/Controllers/AgentControllers/PaymentsController.cs
hasmikharutyunyan: In previous PR was suggestion `pay-by-new-card`. And for me no difference because as I understand we have not separate entities new card and saved card.

Api/Controllers/AgentControllers/InternalBookingsController.cs
hasmikharutyunyan: `for-notify` ? Or what you suggest?

kirillta: Leave as it is

Api/Controllers/AgentControllers/PaymentsController.cs
kirillta: We have schemes where an entity goes before an action. `pay-by-new-card` violates the rule

kirillta (approved):

PR #728, opened 2021-06-09
Feature/aa 303 add notifications for agency management events

AlexeyShok:
Added notification for master agent for changing the status of the agency's activity

### DIFF
--- a/Api/Infrastructure/Options/AgencyManagementMailOptions.cs
+++ b/Api/Infrastructure/Options/AgencyManagementMailOptions.cs
@@ -1,0 +1,7 @@
+﻿namespace HappyTravel.Edo.Api.Infrastructure.Options
+{
+    public class AgencyManagementMailOptions
+    {
+        public string AgencyActivityChangedTemplateId { get; set; }
+    }
+}
--- a/Api/Infrastructure/ServiceCollectionExtensions.cs
+++ b/Api/Infrastructure/ServiceCollectionExtensions.cs
@@ -278,7 +278,7 @@ namespace HappyTravel.Edo.Api.Infrastructure
 
             #region tag processing options
 
-var tagProcessingOptions = vaultClient.Get(configuration["Edo:TagProcessing:Options"]).GetAwaiter().GetResult();
+            var tagProcessingOptions = vaultClient.Get(configuration["Edo:TagProcessing:Options"]).GetAwaiter().GetResult();
             services.Configure<TagProcessingOptions>(options =>
             {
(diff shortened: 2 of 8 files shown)


Api/Infrastructure/ServiceCollectionExtensions.cs
```diff
@@ -269,12 +269,16 @@ public static class ServiceCollectionExtensions
...
+                options.AgencyActivityChangedTemplateId = agencyActivityChangedId;
+            });
             #endregion
-            
+
             #region tag processing options
-            
-            var tagProcessingOptions = vaultClient.Get(configuration["Edo:TagProcessing:Options"]).GetAwaiter().GetResult();
+
+var tagProcessingOptions = vaultClient.Get(configuration["Edo:TagProcessing:Options"]).GetAwaiter().GetResult();
```
oleg1std, line 281: formattin

AlexeyShok: Fixed

Api/Infrastructure/Options/AgencyManagementMailOptions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace HappyTravel.Edo.Api.Infrastructure.Options
+{
+    public class AgencyManagementMailOptions
```
kirillta, line 3: Probably we didn't call it that way, but I think the name should be _Mailing_ options

HappyTravel.Edo.Notifications/Enums/NotificationTypes.cs
```diff
@@ -20,7 +20,7 @@ public enum NotificationTypes
         ChildAgencyInvitation = 12,
         AgentSuccessfulRegistration = 13,
         ChildAgencySuccessfulRegistration = 14,
-        AgencyManagement = 15,
+        AgencyActivityChanged = 15,
```
kirillta, line 23: Why?

Api/Infrastructure/Options/AgencyManagementMailOptions.cs
AlexeyShok: Renamed

HappyTravel.Edo.Notifications/Enums/NotificationTypes.cs
AlexeyShok: We have our own email template for each event. Therefore, each event must have its own value in the enumeration. For counterparty, for example, there are already two such events: changing the verification status and activating / deactivating counterparty. So I renamed this notification type to be more specific.

kirillta (approved):

PR #735, opened 2021-06-15
Feature/add admin endpoint to fix htid

DFSko:


Api/AdministratorServices/FixHtIdService.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        
+        
+        public async Task FillEmptyHtIds()
+        {
+            var bookingsWithEmptyHtId = await _context.Bookings
+                .Where(b => string.IsNullOrEmpty(b.HtId))
+                .ToListAsync();
+
+            var cached = new Dictionary<(Suppliers, string), Result<Accommodation, ProblemDetails>>();
```
mingazhev, line 32: accommodationsCache

mingazhev (approved):
With a small comment

Api/AdministratorServices/FixHtIdService.cs
DFSko: Fixed

PR #737, opened 2021-06-15
Feature/aa 309 add notification about change counterparty verification status

AlexeyShok:
Added notifications for changing activation status of counterparty. Added notifications for changing counterparty vefification state. Fixed unit tests


Api/Models/Mailing/CounterpartyActivityChangedData.cs
```diff
@@ -0,0 +1,11 @@
+﻿using HappyTravel.Edo.Api.Models.Management.Enums;
+
+namespace HappyTravel.Edo.Api.Models.Mailing
+{
+    public class CounterpartyActivityChangedData : DataWithCompanyInfo
```
kirillta, line 5: Could this be `CounterpartyStatusChangedData` or `CounterpartyActivityStatusChangedData` to correspond DB field name?

Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -179,6 +204,27 @@ async Task ChangeCounterpartyAccountsActivityStatus()
...
+
+            async Task<Result> SendNotificationToMaster()
+            {
+                var (_, isFailure, master, error) = await GetRootAgencyMasterAgent(counterparty.Id);
+                if (isFailure)
+                    return Result.Failure(error);
+
+                var messageData = new CounterpartyActivityChangedData
+                {
+                    AgentName = $"{master.FirstName} {master.LastName}",
```
kirillta, line 217: We have so many formatters like this, probably they have to be implemented as a separate method. Just an advise

Api/Models/Mailing/CounterpartyActivityChangedData.cs
AlexeyShok: This field is named IsActive in the database.

AlexeyShok: Renamed

Api/AdministratorServices/CounterpartyManagementService.cs
AlexeyShok: Added get property to model

kirillta (approved):

PR #741, opened 2021-06-18
Add agency roles

konstyefremov:


HappyTravel.Edo.Data/Agents/AgentAgencyRelation.cs
```diff
@@ -5,6 +5,8 @@ namespace HappyTravel.Edo.Data.Agents
     public class AgentAgencyRelation
     {
         public int AgentId { get; set; }
+        public Role Role { get; set; }
+        public int RoleId { get; set; }
```
konstyefremov, line 9: this probably should be nullable

Raibo (approved):
I approve even without table or dictionary of roles, because it can be added later

PR #757, opened 2021-06-23
Crud for agent roles

Raibo:
https://github.com/happy-travel/agent-app-project/issues/329


Raibo:
I did not test it yet, because the table is empty, and I don't want to mess with Ids. I will test this after the data will be there.

Api/AdministratorServices/AgentRolesService.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Threading.Tasks;
+using FluentValidation;
+using HappyTravel.Edo.Api.AdministratorServices.Models;
+using HappyTravel.Edo.Api.Extensions;
+using HappyTravel.Edo.Api.Infrastructure;
+using HappyTravel.Edo.Data.Agents;
+
+namespace HappyTravel.Edo.Api.AdministratorServices
+{
+    public class AgentRolesService : IAgentRolesService
```
mingazhev, line 16: AgentRolesManagementService will be more clear, I think

Api/AdministratorServices/AgentRolesService.cs
```diff
@@ -0,0 +1,107 @@
...
+                _context.Update(role);
+                return _context.SaveChangesAsync();
+            }
+        }
+
+
+        public async Task<Result> Delete(int roleId)
+        {
+            return await Get(roleId)
+                .Ensure(IsUnused, "This role is in use and connot be deleted")
```
mingazhev, line 75: Cannot

Api/AdministratorServices/AgentRolesService.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            Task Add()
+            {
+                _context.AgentRoles.Add(roleInfo.ToAgentRole());
+                return _context.SaveChangesAsync();
+            }
+        }
+
+
+        public async Task<Result> Edit(int roleId, AgentRoleInfo roleInfo)
```
konstyefremov, line 48: Do we need to pass "int roleId" here if it's present inside "AgentRoleInfo roleInfo"?

Api/AdministratorServices/AgentRolesService.cs
```diff
@@ -0,0 +1,107 @@
...
+
+
+        public async Task<Result> Delete(int roleId)
+        {
+            return await Get(roleId)
+                .Ensure(IsUnused, "This role is in use and connot be deleted")
+                .Tap(Delete);
+
+
+            async Task<bool> IsUnused(AgentRole _)
```
konstyefremov, line 79: "AgentRole _" looks like can be safely removed.

Api/AdministratorServices/AgentRolesService.cs
```diff
@@ -0,0 +1,107 @@
...
+                return _context.SaveChangesAsync();
+            }
+        }
+
+
+        private Result Validate(AgentRoleInfo roleInfo)
+            => GenericValidator<AgentRoleInfo>.Validate(v =>
+                {
+                    v.RuleFor(r => r.Name).NotEmpty();
+                    v.RuleFor(r => r.Permissions.ToList()).NotEmpty();
```
konstyefremov, line 95: Is it appropriate to use .ToList() here?

Api/Controllers/AdministratorControllers/AgentRolesController.cs
```diff
@@ -0,0 +1,73 @@
...
+    public class AgentRolesController : BaseController
+    {
+        public AgentRolesController(IAgentRolesService agentRolesService)
+        {
+            _agentRolesService = agentRolesService;
+        }
+
+
+        /// <summary>
+        ///     Gets all pssible agent roles
```
konstyefremov, line 25: possible

Api/AdministratorServices/AgentRolesService.cs
Raibo: Yes, we absolutely need, for two reasons:
1. API consistency, we gain access to an entity using it's Id in the endpoint path
2. Id in AgentRoleInfo instances are ignored. It is there only for responses, not requests.

Api/AdministratorServices/AgentRolesService.cs
Raibo: done

Api/AdministratorServices/AgentRolesService.cs
Raibo: This is due to FunctionalExtensions looks better when signature fits

Api/AdministratorServices/AgentRolesService.cs
Raibo: fixed

Raibo:
fixed and tested

mingazhev (approved):

PR #758, opened 2021-06-23
Added string enum converter for enums

AlexeyShok:
Added string enum converter for enums


DFSko (approved):

mingazhev (changes requested):
Need to use EnumFormatters for e-mails

mingazhev (approved):

PR #762, opened 2021-06-24
Feature/aa 332 add admin roles migration

AlexeyShok:
Added AdminRole model. Added migration for create table and save roles for existing admins. ReportGeneration admin permission split into 3 permissions


mingazhev (approved):
Needs @kirillta approve

HappyTravel.Edo.Data/EdoContext.cs
```diff
@@ -98,6 +95,7 @@ public EdoContext(DbContextOptions<EdoContext> options) : base(options)
         
         public DbSet<Discount> Discounts { get; set; }
         public DbSet<AgentRole> AgentRoles { get; set; }
+        public DbSet<AdministratorRole> AdministratorRoles { get; set; }
```
konstyefremov, line 98: Should be virtual if in the future it will be unit tested.

AlexeyShok: Fixed

PR #363, opened 2020-12-10
Bugfix/nijo 1097 booking flow failed if supplier reject booking

DFSko:

PR #770, opened 2021-06-28
Returns available agent roles

konstyefremov:
resolves https://github.com/happy-travel/agent-app-project/issues/363


mingazhev (approved):
After changes

PR #771, opened 2021-06-28
Changed permission checks to use permissions instead of roles (temp r…

mingazhev:
…evert back)

### DIFF
--- a/Api/Services/Agents/PermissionChecker.cs
+++ b/Api/Services/Agents/PermissionChecker.cs
@@ -28,39 +28,55 @@ namespace HappyTravel.Edo.Api.Services.Agents
                 agent.AgencyId.ToString(),
                 agent.AgentId.ToString());
             
-            var storedPermissions = await _flow.GetOrSetAsync(
-                key: key, 
+            var storedPermissions = await _flow.GetOrSetAsync(key: key, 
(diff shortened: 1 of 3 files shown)


Api/Services/Agents/PermissionChecker.cs
```diff
@@ -28,39 +28,55 @@ public async ValueTask<Result> CheckInAgencyPermission(AgentContext agent, InAge
...
+            //         .Where(r => r.AgencyId == agencyId)
+            //         .Select(r => r.AgentRoleIds)
+            //         .SingleOrDefaultAsync();
+            //
+            //     if (roleIds == default)
+            //         return default;
+            //         
+            //     var permissions = await _context.AgentRoles
+            //         .Where(r => roleIds.Contains(r.Id))
+            //         .Select(r => r.Permissions)
```
AlexeyShok, line 75: Since the rights in the roles can be duplicated, we need to add the removal of duplicate rights. May be .Distinct()?

AlexeyShok (approved):

Api/Services/Agents/PermissionChecker.cs
mingazhev: Maybe, but not mandatory. If one of the permission will match, the LINQ operation will end executing

PR #774, opened 2021-06-29
added prometheus booking counter

DFSko:


mingazhev (commented):
Not sure that this counter is needed now, at least as a single

PR #776, opened 2021-06-29
Add roles to agent invitation

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/365

The invitation data stored in UserDescriptionInfo will also contain RoleIds - ids of assigned roles.
As this class is used for both agent's and admin's registration, the name will be more generic - just RoleIds, not AgentRoleIds or AdminRoleIds.

This data is provided by the front and then stored inside JSON for later use.

### DIFF
--- a/Api/Models/Users/UserDescriptionInfo.cs
+++ b/Api/Models/Users/UserDescriptionInfo.cs
@@ -7,13 +7,14 @@ namespace HappyTravel.Edo.Api.Models.Users
     {
         [JsonConstructor]
         public UserDescriptionInfo(string title, string firstName, string lastName,
-            string position, string email)
+            string position, string email, int[] roleIds)
         {
             Title = title;
             FirstName = firstName;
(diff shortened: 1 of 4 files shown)


Api/Models/Users/UserDescriptionInfo.cs
```diff
@@ -7,13 +7,14 @@ namespace HappyTravel.Edo.Api.Models.Users
...
         public UserDescriptionInfo(string title, string firstName, string lastName,
-            string position, string email)
+            string position, string email, int[] roleIds)
         {
             Title = title;
             FirstName = firstName;
             LastName = lastName;
             Position = position;
             Email = email;
+            RoleIds = roleIds;
```
Raibo, line 17: Maybe we should use role names instead of Ids, because we can free the users from bothering with role Ids altogether.

mingazhev (approved):

PR #788, opened 2021-07-05
fixed multiroom price calculation

DFSko:
resolves happy-travel/agent-app-project#191


mingazhev (approved):
Fine for me but let's see other reviews

kirillta (approved):

mingazhev (approved):

PR #793, opened 2021-07-05
Don't try to send a supplier booking request for offline bookings payed by credit card

Raibo:
https://github.com/happy-travel/agent-app-project/issues/392

If we deal with an offline booking, then it is confirmed upon creating.
If we then try to pay for it by credit card, an attempt to confirm it again is made, because credit card bookings are not normally confirmed upon creation, only when payed. 

We need to take into account that the booking could be of offline type before and skip confirmation if a booking is already confirmed.

### DIFF
--- a/Api/Services/Accommodations/Bookings/BookingExecution/Flows/BankCreditCardBookingFlow.cs
+++ b/Api/Services/Accommodations/Bookings/BookingExecution/Flows/BankCreditCardBookingFlow.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
 using HappyTravel.Edo.Api.Infrastructure;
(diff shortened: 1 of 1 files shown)


Api/Services/Accommodations/Bookings/BookingExecution/Flows/BankCreditCardBookingFlow.cs
```diff
@@ -80,7 +83,7 @@ public async Task<Result<AccommodationBookingInfo>> Finalize(string referenceCod
                 .Check(CheckBookingIsPaid)
                 .CheckIf(IsDeadlinePassed, CaptureMoney)
                 .Check(GenerateInvoice)
-                .Bind(SendSupplierRequest)
+                .Bind(SendSupplierRequestIfNeeded)
```
mingazhev, line 86: This approach seems totally incorrect, not only because of sending additional supplier requests, most of other methods should not be used too.
Please see `refcode/{referenceCode}/pay-with-credit-card` endpoint to see the actual method.
If it is not called from frontend, this is frontend issue.

mingazhev (changes requested):

PR #795, opened 2021-07-06
added HappyTravel.Telemetry package

DFSko:


mingazhev (approved):
Need a new version of package and merge after code freeze end only

PR #813, opened 2021-07-09
Added url generate service

AlexeyShok:
Added service for URL generation. Using AES encryption for generate URL. Endpoints for hotel confirmation page now receive encrypted reference code. Secrets added to Vault.

### DIFF
--- a/Api/Services/PropertyOwners/UrlGenerationService.cs
+++ b/Api/Services/PropertyOwners/UrlGenerationService.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.IO;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace HappyTravel.Edo.Api.Services.PropertyOwners
+{
+    public class UrlGenerationService : IUrlGenerationService
+    {
+        public string Generate(string referenceCode)
+        {
+            var variablePartOfUrl = GenerateFromString(referenceCode);
+
+            return $"{constantPartOfUrl}/{variablePartOfUrl}";
+        }
+
+
+        private static string GenerateFromString(string stringToEncrypt)
+        //    => Convert.ToBase64String(Encoding.ASCII.GetBytes(referenceCode));
+        {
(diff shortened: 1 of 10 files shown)


Api/Services/PropertyOwners/UrlGenerationService.cs
```diff
@@ -0,0 +1,132 @@
...
+            if (cipherText == null || cipherText.Length <= 0)
+                throw new ArgumentNullException("cipherText");
+            if (Key == null || Key.Length <= 0)
+                throw new ArgumentNullException("Key");
+            if (IV == null || IV.Length <= 0)
+                throw new ArgumentNullException("IV");
+
+            // Declare the string used to hold
+            // the decrypted text.
+            string plaintext = null;
```
kirillta, line 94: `plainText`

Api/Services/PropertyOwners/UrlGenerationService.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public string Generate(string referenceCode)
+        {
+            var variablePartOfUrl = Encrypt(referenceCode);
+
+            return $"{_urlGenerationOptions.ConfirmationPageUrl}/{variablePartOfUrl}";
+        }
+
+
+        public string Decrypt(string encryptedString)
```
kirillta, line 25: `Generate`/`Decrypt` isn't look like a pair actions. Besides we pass only a reference code to `Generate`. I suggest to rename either `Decrypt` to `ReadReferenceCode`, or `Generate` to `Encrypt`

kirillta (approved):

Api/Services/PropertyOwners/UrlGenerationService.cs
AlexeyShok: Renamed

Api/Services/PropertyOwners/UrlGenerationService.cs
AlexeyShok: Renamed to ReadReferenceCode

PR #824, opened 2021-07-20
Validate discounts from discounts and markups, add unit tests

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/355

### DIFF
--- a/Api/AdministratorServices/AgencyDiscountManagementService.cs
+++ b/Api/AdministratorServices/AgencyDiscountManagementService.cs
@@ -6,7 +6,9 @@ using CSharpFunctionalExtensions;
 using HappyTravel.Edo.Api.AdministratorServices.Models;
 using HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions;
 using HappyTravel.Edo.Api.Models.Management.AuditEvents;
+using HappyTravel.Edo.Api.Services;
 using HappyTravel.Edo.Api.Services.Management;
+using HappyTravel.Edo.Api.Services.Markups.Templates;
 using HappyTravel.Edo.Common.Enums;
 using HappyTravel.Edo.Common.Enums.Markup;
 using HappyTravel.Edo.Data;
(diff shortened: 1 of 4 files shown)


AlexeyShok (approved):

Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -83,6 +89,24 @@ async Task<Result> ValidateAgency()
                     : Result.Failure($"Could not find an agency with id {agencyId}");
 
 
+            async Task<Result> DiscountDoesntExceedMarkups()
```
kirillta, line 92: I suggest `DoesDiscountExceedsMarkups`

Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -83,6 +89,24 @@ async Task<Result> ValidateAgency()
...
+            {
+                var markupPolicy = await _context.MarkupPolicies.SingleOrDefaultAsync(x => x.Id == createDiscountRequest.TargetMarkupId);
+                var markupFunction = _templateService.CreateFunction(markupPolicy.Id, markupPolicy.TemplateSettings);
+                
+                var allDiscounts = await _context.Discounts
+                    .Where(x => x.TargetPolicyId == markupPolicy.Id)
+                    .Where(d => d.TargetAgencyId == agencyId)
```
kirillta, line 99: Why we check only an agency?

Api/AdministratorServices/AgencyDiscountManagementService.cs
konstyefremov: I'd disagree here. This check is used within the validation block.

return ValidatePercent(createDiscountRequest.DiscountPercent)
                .Bind(ValidateTargetMarkup)
                .Bind(ValidateAgency)
                .Bind(DiscountDoesntExceedMarkups)

If I rename it to `DoesDiscountExceedsMarkups` it will not fit this validation flow anymore.

Api/AdministratorServices/AgencyDiscountManagementService.cs
konstyefremov: Because discounts are linked only to specific markups and specific agencies.

Api/AdministratorServices/AgencyDiscountManagementService.cs
kirillta: Either that, either 'DoesNotDiscount...'. I mean, the current name don't tell us what the method check something

PR #436, opened 2021-01-21
fixed error on sending invite with empty email

DFSko:

PR #831, opened 2021-07-24
Added sales report

Raibo:
https://github.com/happy-travel/agent-app-project/issues/436


Raibo:
We might want to introduce indexes for `CreationDate` and `CheckOutDate`, since we are pulling data from ever growing table using this columns to filter.

Api/Services/Reports/RecordManagers/SalesBookingsRecordManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+
+        private DateTime GetNonRefundableDate(SalesBookingsReportProjection projection)
+        {
+            var lastPolicy = projection.CancellationPolicies
+                .OrderBy(p => p.FromDate)
+                .FirstOrDefault(p => p.Percentage >= 100d);
+            return lastPolicy?.FromDate ?? projection.CheckInDate;
+        }
```
Raibo, line 95: >for refundable bookings

I am taking `refundable` as if 100% refund penalty (or check in date) is after current datetime.

Api/Services/Reports/RecordManagers/IRecordManager.cs
```diff
@@ -1,10 +1,11 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace HappyTravel.Edo.Api.Services.Reports.RecordManagers
 {
-    public interface IRecordManager<out TProjection>
+    public interface IRecordManager<TProjection>
     {
-        IQueryable<TProjection> Get(DateTime fromDate, DateTime endDate);
+        Task<IEnumerable<TProjection>> Get(DateTime fromDate, DateTime endDate);
```
Raibo, line 9: I am refactoring `RecordManagers` to become asynchronous, because currently DB queries are made synchronously inside `foreach`, which is not good. 
Also, I need to use client evaluation in new report to filter out required bookings.

Api/Services/Reports/RecordManagers/SalesBookingsRecordManager.cs
```diff
@@ -0,0 +1,101 @@
...
+                    .FirstOrDefault()
+                where
+                    (booking.CheckOutDate >= fromDate &&
+                    booking.CheckOutDate < endDate
+                    ||
+                    booking.Created >= fromDate &&
+                    booking.Created < endDate)
+                    &&
+                    (booking.Status == BookingStatuses.Confirmed ||
+                    booking.Status == BookingStatuses.Cancelled && cancellationDate >= booking.DeadlineDate)
```
Raibo, line 41: >Only bookings that are needed to pay should be shown. This can include confirmed bookings or bookings that were cancelled after deadline

This cuts most of bookings compared to full bookings report

Api/Services/Reports/RecordManagers/FullBookingsRecordManager.cs
```diff
@@ -14,41 +17,42 @@ public FullBookingsRecordManager(EdoContext context)
         }
         
         
-        public IQueryable<FullBookingsReportProjection> Get(DateTime fromDate, DateTime endDate)
+        public async Task<IEnumerable<FullBookingsReportProjection>> Get(DateTime fromDate, DateTime endDate)
```
kirillta, line 20: Same here

Api/Services/Reports/ReportService.cs
```diff
@@ -58,17 +59,32 @@ public async Task<Result<Stream>> GetFullBookingsReport(DateTime fromDate, DateT
...
+
+            Task<IEnumerable<FullBookingsReportProjection>> GetRecords()
                 => GetRecords<FullBookingsReportProjection>(from, end);
         }
 
 
+        public async Task<Result<Stream>> GetSalesBookingsReport(DateTime fromDate, DateTime endDate)
+        {
+            var from = fromDate.Date;
+            var end = endDate.Date.AddDays(1).AddTicks(-1);
```
kirillta, line 76: `from` and `to`

Api/Services/Reports/RecordManagers/AgencyWiseRecordManager.cs
```diff
@@ -13,13 +16,13 @@ public AgencyWiseRecordManager(EdoContext context)
         }
         
         
-        public IQueryable<AgencyWiseRecordProjection> Get(DateTime fromDate, DateTime endDate)
+        public async Task<IEnumerable<AgencyWiseRecordProjection>> Get(DateTime fromDate, DateTime endDate)
```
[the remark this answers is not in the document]
Raibo: I can rename.
Or cast AsQueriable, but that would not make much sense.

But I can't make this back sync as it was before, both because it caused sync DB data retrieval and because I need client evaluation for new report.

Api/Services/Reports/ReportService.cs
Raibo: I just copied how it was for all other reports. 
Again got into a refactoring trouble.

mingazhev (approved):
Let's merge as is and then review again with no hurry.

Api/Services/Reports/RecordManagers/AgencyWiseRecordManager.cs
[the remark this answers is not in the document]
Raibo: And so I renamed it all from "projection" to "data"

PR #836, opened 2021-07-26
Feature/AA 409 Pending Supplier Reference and Confirmed Bookings Reports

konstyefremov:
https://github.com/happy-travel/agent-app-project/issues/409

Added two reports. Need to check if I understood the applied filters correctly.

Also, these reports mostly copy each other with the only difference in the 'where' block. What would be a better way to keep it DRY?


mingazhev:
Please fix the conflicts

mingazhev (commented):
Need conflicts fix

kirillta (approved):

Api/Infrastructure/ServiceCollectionExtensions.cs
```diff
@@ -778,6 +778,16 @@ public static IServiceCollection AddServices(this IServiceCollection services)
             services.AddTransient<IRecordManager<SupplierWiseRecordData>, SupplierWiseRecordsManager>();
             services.AddTransient<IRecordManager<FullBookingsReportData>, FullBookingsRecordManager>();
             services.AddTransient<IRecordManager<SalesBookingsReportData>, SalesBookingsRecordManager>();
+            services.AddTransient<IConverter<AgencyWiseRecordData, AgencyWiseReportRow>, AgencyWiseRecordDataConverter>();
```
mingazhev, line 781: Consider moving all report needed registrations to a separate method

mingazhev (approved):
With a small comment

PR #839, opened 2021-07-28
Hotfix/FixNotChargedBookings

oleg1std:
- remove checkInDates checks from the charge method
- add a migration to set Confirmed status for the not charged bookings


Raibo (approved):
I guess Ids in migration are for production, while bookings with similar Ids at dev will be casualties. If it is ok, then I guess ok.

PR #840, opened 2021-07-28
Added check for cancellation policies into BookingEvaluationService

Raibo:


Api/Services/Accommodations/Availability/Steps/BookingEvaluation/BookingEvaluationService.cs
```diff
@@ -197,8 +198,24 @@ Task SaveToCache(DataWithMarkup<EdoContracts.Accommodations.RoomContractSetAvail
...
+
+            Result<Unit, ProblemDetails> CheckCancellationPolicies(RoomContractSetAvailability? availability)
+            {
+                if (availability is null)
+                    return Unit.Instance;
+
+                var availabilityValue = availability.Value;
+                var deadline = availabilityValue.RoomContractSet.Deadline;
+
+                var invalid = deadline is null || deadline.Policies.Any(p => p.Percentage == 0d);
```
mingazhev, line 211: isInvalid

Api/Services/Accommodations/Availability/Steps/BookingEvaluation/BookingEvaluationService.cs
```diff
@@ -197,8 +198,24 @@ Task SaveToCache(DataWithMarkup<EdoContracts.Accommodations.RoomContractSetAvail
...
+
+                return invalid
```
mingazhev, line 213: Please log error in such case

mingazhev: And maybe we need a comment why we do that

mingazhev (approved):

PR #842, opened 2021-07-29
Bugfix/aa 355 dont discount more than cost of booking

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/355#issuecomment-887850503


Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -62,7 +62,7 @@ public Task<Result> Add(int agencyId, CreateDiscountRequest createDiscountReques
             return ValidatePercent(createDiscountRequest.DiscountPercent)
                 .Bind(ValidateTargetMarkup)
                 .Bind(ValidateAgency)
-                .Bind(ValidateDiscountSize)
+                .Bind(DiscountsDontExceedMarkups)
```
DFSko, line 65: Maybe name should be a verb

konstyefremov: > Maybe name should be a verb

We previously discussed this with Kirill. This name is a better fit.

mingazhev: Why not using `Check` instead of `Bind`?

Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -182,6 +183,32 @@ Task Update(Discount discount)
 
             Task<Result> WriteAuditLog(Discount _)
                 => _managementAuditService.Write(ManagementEventType.DiscountEdit, new DiscountActivityStateEventData(agencyId, newActivityState));
+
+
+            async Task<Result<Discount>> DiscountDoesntExceedMarkups(Discount discount)
+            {
+                // we need additional check for activated discounts
```
mingazhev, line 190: Please start comments from a capital letter

Api/AdministratorServices/AgencyDiscountManagementService.cs
konstyefremov: It will cause a type mismatch and the whole chain needs to be fixed. 
Already discussed.

Api/AdministratorServices/AgencyDiscountManagementService.cs
konstyefremov: Done

mingazhev (approved):

PR #852, opened 2021-08-04
Check default guids for room contract sets

konstyefremov:


Api/Services/Accommodations/Availability/DeadlineService.cs
```diff
@@ -72,7 +75,7 @@ public class DeadlineService : IDeadlineService
                 {
                     var selectedRoom = a.RoomContractSets?.SingleOrDefault(r => r.Id == roomContractSetId);
 
-                    if (selectedRoom is not null)
+                    if (selectedRoom is not null && !selectedRoom.Id.Equals(default))
```
DFSko, line 78: default guid is Guid.Empty
maybe check selectedRoom.Id != Guid.Empty?

mingazhev: default sounds better for me

mingazhev (approved):

PR #860, opened 2021-08-10
Refactoring and fixing PayableToSupplierOrHotel calculation

konstyefremov:
https://github.com/happy-travel/agent-app-project/issues/436#issuecomment-889117211

### DIFF
--- a/Api/Services/Reports/ReportService.cs
+++ b/Api/Services/Reports/ReportService.cs
@@ -76,7 +76,7 @@ namespace HappyTravel.Edo.Api.Services.Reports
             var to = endDate.Date.AddDays(1);
 
             return await Validate(from, to)
-                .Ensure(NoFutureDate, "Cannot query future dates for this report")
+                .Ensure(DatesAreNotInFuture, "Cannot query future dates for this report")
                 .Map(GetRecords)
                 .Bind(Generate<SalesBookingsReportData, SalesBookingsReportRow>);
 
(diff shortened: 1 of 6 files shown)


Api/Services/Reports/ReportService.cs
```diff
@@ -74,12 +76,20 @@ public async Task<Result<Stream>> GetSalesBookingsReport(DateTime fromDate, Date
...
+                .Ensure(NoFutureDate, "Cannot query future dates for this report")
                 .Map(GetRecords)
                 .Bind(Generate<SalesBookingsReportData, SalesBookingsReportRow>);
 
 
             Task<IEnumerable<SalesBookingsReportData>> GetRecords()
                 => GetRecords<SalesBookingsReportData>(from, to);
+
+
+            bool NoFutureDate()
```
mingazhev, line 88: DatesAreNotInFuture

mingazhev: + this can be included into "Validate" mtehod

mingazhev (approved):
With small comments

mingazhev (changes requested):
Need to optimize the query not to load all the bookings

Api/Services/Reports/ReportService.cs
konstyefremov: Done

mingazhev (approved):

PR #413, opened 2021-01-11
added better invitations info, endpoints for getting accepted invitat…

DFSko:
…ions list, endpoint for disable invitation

PR #879, opened 2021-08-17
Remove InAgencyPermissions column from AgentAgencyRelation

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/366
Resolves https://github.com/happy-travel/agent-app-project/issues/413


Api/Services/Agents/AgentRegistrationService.cs
```diff
@@ -140,7 +140,6 @@ private Task AddAgentAgencyRelation(Agent agent, AgentAgencyRelationTypes relati
             {
                 AgentId = agent.Id,
                 Type = relationType,
-                InAgencyPermissions = permissions,
```
mingazhev, line 143: Should not you delete a method parameter `permissions` then?

konstyefremov: Removed this and checked again the code for remnants of old code.

konstyefremov:
Also deleted Permissions Sets file.

konstyefremov:
Also removed Skip from two unit tests on Permission Checks.

mingazhev (approved):
To merge after code freeze only (!)

PR #886, opened 2021-08-19
Feature/aa 409 add additional data to analytics events

konstyefremov:


Api/Services/Analytics/AnalyticsService.cs
```diff
@@ -1,23 +1,23 @@
...
-using HappyTravel.DataFormatters;
 using HappyTravel.MapperContracts.Internal.Mappings.Internals;
 using HappyTravel.MapperContracts.Public.Accommodations.Internals;
 using Accommodation = HappyTravel.MapperContracts.Public.Accommodations.Accommodation;
 
-namespace HappyTravel.Edo.Api.Services.Accommodations.Availability
+namespace HappyTravel.Edo.Api.Services.Analytics
 {
-    public class AvailabilityAnalyticsService
+    public class AnalyticsService
```
mingazhev, line 18: This is too common name for the service. We already have generic AnalyticsService unbound from the business logic.
If you'd like to use this service for availability + booking events, let's name it accordingly,
like `BookingAnalyticsService` which will include availability searches and bookings

Api/Services/Analytics/AnalyticsService.cs
```diff
@@ -81,6 +81,30 @@ public void LogAccommodationAvailabilityRequested(in Accommodation accommodation
         }
         
         
+        public void LogBookingStatusChange(Booking booking,
```
mingazhev, line 84: This event looks too generic as I see.
It is not used for all status changes (which is correct), only for 2 of them. Why not using 2 separate events?
Saving this in a single event with unclear semantics will make reporting and analyzing harder as I see.

Api/Services/Analytics/AnalyticsService.cs
konstyefremov: Fixed

Api/Services/Analytics/AnalyticsService.cs
konstyefremov: Fixed.

Api/Models/Analytics/AgentAnalyticsInfo.cs
```diff
@@ -0,0 +1,12 @@
+namespace HappyTravel.Edo.Api.Models.Analytics
+{
+    public readonly struct AgentAnalyticsInfo
+    {
+        public AgentAnalyticsInfo(string counterpartyName)
+        {
+            CounterpartyName = counterpartyName;
+        }
+        
+        public string CounterpartyName { get; }
```
DFSko, line 10: Do you really need a struct with only one property?

konstyefremov: Yes. We discussed this with Vadim on Thursday. It most likely will be extended later with more properties.

Api/Services/Accommodations/Bookings/Management/BookingRecordsUpdater.cs
```diff
@@ -158,22 +164,34 @@ async Task<Result> SendInvoice(AccommodationBookingInfo bookingInfo)
...
+
+
             void WriteFailureLog(string error) 
                 => _logger.LogBookingConfirmationFailure(booking.ReferenceCode, error);
         }
         
         
         private Task<Result> ProcessCancellation(Booking booking, DateTime cancellationDate, ApiCaller user)
         {
+            var counterParty = _context.Counterparties
```
mingazhev, line 178: this is a single word, not two. `counterparty`

Api/Services/Analytics/PassengersInfo.cs
```diff
@@ -0,0 +1,29 @@
+using System.Collections.Generic;
+using System.Linq;
+using HappyTravel.Edo.Data.Bookings;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+
+namespace HappyTravel.Edo.Api.Services.Analytics
+{
+    public readonly struct PassengersInfo
+    {
+        public PassengersInfo(List<BookedRoom> rooms)
```
mingazhev, line 10: You should avoid logic in constructors. Consider using static factory method to create the struct in this case

mingazhev (approved):
With small comments. May be merged after fixing

Api/Services/Analytics/PassengersInfo.cs
konstyefremov: replaced with static factory method

Api/Services/Accommodations/Bookings/Management/BookingRecordsUpdater.cs
konstyefremov: fixed

PR #921, opened 2021-08-31
Generate incoive after money is charged from account

Raibo:
https://github.com/happy-travel/agent-app-project/issues/502

Should be okay, but I did not check (can't pay with account with my agency on dev yet).
If it is obvious that this will work (for credit card flow it works), then we can merge as is.


mingazhev (approved):
merge after being tested

Api/Services/Accommodations/Bookings/Documents/BookingDocumentsService.cs
```diff
@@ -195,7 +195,7 @@ public async Task<Result<(DocumentRegistrationInfo RegistrationInfo, PaymentRece
         public async Task<Result<(DocumentRegistrationInfo RegistrationInfo, BookingInvoiceData Data)>> GetActualInvoice(Booking booking)
         {
             var lastInvoice = (await _invoiceService.Get<BookingInvoiceData>(ServiceTypes.HTL, ServiceSource.Internal, booking.ReferenceCode))
-                .OrderByDescending(i => i.Metadata.Date)
+                .OrderBy(i => i.Metadata.Date)
                 .LastOrDefault();
```
Raibo, line 199: This was probably an error.
If we sort by date desc and then take last one, this would end up taking the first (oldest) invoice.

Raibo:
@mingazhev Have a look one more time please

mingazhev (approved):

PR #531, opened 2021-03-15
added migration to fill ContractKindColumn in Counterparty table

hasmikharutyunyan:

PR #925, opened 2021-09-01
Avoid trying to create Result with empty error

Raibo:
https://github.com/happy-travel/agent-app-project/issues/531


mingazhev (approved):
fine

PR #933, opened 2021-09-02
Remove dependency on UserDescriptionInfo in SendAgentInvitationRequest

konstyefremov:
https://github.com/happy-travel/agent-app-project/issues/453


konstyefremov:
Merge only after changes on the frontend

AlexeyShok (commented):
Where is the email validation?

konstyefremov:
The request will be changed later on the front. For now we need only email check.

PR #538, opened 2021-03-17
implemented notification center

DFSko:

PR #935, opened 2021-09-04
Fix wrong percentage in booking cancellation

Raibo:
https://github.com/happy-travel/agent-app-project/issues/538


Raibo:
Did not check explicitly, but we have cancellation policies ranges 0..100.
All the weird prices are 99 times bigger or 49 times bigger than the actual price, and so on.
Also all the weird bookings are cancelled.

mingazhev (approved):

PR #567, opened 2021-03-25
added settings for only direct contract search

DFSko:
added settings for only direct contract search

p.4 in happy-travel/agent-app-project#88

PR #943, opened 2021-09-09
Table and management service for balance notification settings

Raibo:
https://github.com/happy-travel/agent-app-project/issues/567

### DIFF
--- a/Api/Models/Agencies/BalanceNotificationSettingInfo.cs
+++ b/Api/Models/Agencies/BalanceNotificationSettingInfo.cs
@@ -1,0 +1,8 @@
+﻿namespace HappyTravel.Edo.Api.Models.Agencies
+{
+    public readonly struct BalanceNotificationSettingInfo
+    {
+        public int AccountId { get; init; }
+        public int[] Thresholds { get; init; }
+    }
+}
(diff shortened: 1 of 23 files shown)


HappyTravel.Edo.Data/Agents/BalanceNotificationSetting.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace HappyTravel.Edo.Data.Agents
+{
+    public class BalanceNotificationSetting
+    {
+        public int Id { get; set; }
+        public int AccountId { get; set; }
```
mingazhev, line 6: AgencyAccountId maybe? To be more clear

Api/Models/Agencies/BalanceNotificationSettingInfo.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace HappyTravel.Edo.Api.Models.Agencies
+{
+    public readonly struct BalanceNotificationSettingInfo
+    {
+        public int AccountId { get; init; }
```
mingazhev, line 5: Is accountId needed in this model?

mingazhev:
Added some comments

Api/Models/Agencies/BalanceNotificationSettingInfo.cs
Raibo: Technically no, but I just wanted the model to be self-sufficient and unambiguous, for this I added AccountId.

mingazhev (approved):

Issue #568, opened 2021-03-26
Added markups are not implemented


sergeyf-QA:
**How to reproduce:**
- log in as any user
- try to find any booking e.g.
Jumeirah Emirates Towers, Dubai, The United Arab Emirates
29-04-2021 – 30-04-2021
1 Adult • 0 Children • 1 Room
United Kingdom of Great Britain and Northern
United Kingdom of Great Britain and Northern
- Jumeirah Emirates Towers has price 205,48
- add Agent markup for this user
- wait 5min for back-end cache
- try to find this hotel again

**Actual result:**
- price is not changed according to markup

**Expected result:**
- I should see changed price after new markup implementation

DFSko:
try to wait about 15 minutes

sergeyf-QA:
**Summary:**
there is no issue, markup did not work for user with 217 Id because markups were disabled for his agency.

So, OK.

PR #947, opened 2021-09-10
Added notifications for management when account falls below a threshold

Raibo:
happy-travel/agent-app-project#568


Api/Services/Payments/Accounts/AccountPaymentService.cs
```diff
@@ -196,6 +202,16 @@ async Task<Result<Payment>> StorePayment((int accountId, MoneyAmount amount) cha
             Task<Result> ProcessPaymentResults(Payment payment) 
                 => paymentCallbackService.ProcessPaymentChanges(payment);
 
+
+            async Task FillInitialBalance(int accountId)
```
mingazhev, line 206: I suggest to avoid such hacks with filling local variables from the pipeline

mingazhev: Instead of it, you can:
**Variant 1**
1. Return a model like 'AccountInfo' instead of id in `IPaymentCallbackService`. Account info can contain Id and the balance
2. Use this model later in pipe

**Variant 2**
1. Change return type from `IAccountPaymentProcessingService` `Charge` method to return something like ChargeResult, containing AmountBefore or AmountAfter
2. Use this value in notification

Variant 1 is more preferrable for me

mingazhev: And maybe move this logic to account payment processing service at all? Need to think

mingazhev (approved):

PR #950, opened 2021-09-10
Saving booking availability info on booking registration step (3/4)

mingazhev:
This will allow to save all the booking requests and to use simplier methods in the flows.
Tests will be fixed in next PR


Api/Services/Accommodations/Bookings/BookingExecution/BookingRegistrationService.cs
```diff
@@ -41,6 +42,7 @@ public class BookingRegistrationService : IBookingRegistrationService
             var (_, _, booking, _) = await Result.Success()
                 .Map(GetTags)
                 .Map(Create)
+                .Tap(SaveRequestInfo)
```
kirillta, line 45: Should these be all taps?

kirillta (approved):

Api/Services/Accommodations/Bookings/BookingExecution/BookingRegistrationService.cs
mingazhev: Yes, all these functions are not supposed to fail

PR #954, opened 2021-09-13
added payment date to supplier orders tables

DFSko:
resolves happy-travel/agent-app-project#575


mingazhev (approved):
Please check that migration worked as expected after merge

PR #972, opened 2021-09-16
If an agent has bookings the change-agency request return 400

AlexeyShok:
If an agent has bookings, the change-agency request now return 400. Accordingly, the transfer does not take place.
If the source agency has no more agents, it is deactivated. If there are no more active agencies in the source counterparty, it is deactivated.


Api/Services/Management/AgentMovementService.cs
```diff
@@ -12,18 +13,24 @@ namespace HappyTravel.Edo.Api.Services.Management
...
         
         
         public async Task<Result> Move(int agentId, int sourceAgencyId, int targetAgencyId, List<int> roleIds)
         {
             return await ValidateRequest()
+                .Ensure(AgentHasNoBookings, $"Agent {agentId} has bookings in agency {sourceAgencyId}")
                 .Bind(UpdateAgencyRelation)
-                .Bind(WriteLog);
+                .Bind(WriteLog)
+                .Bind(DeactivateAgencyIfNeed)
```
mingazhev, line 32: If needed

Api/Services/Management/AgentMovementService.cs
```diff
@@ -12,18 +13,24 @@ namespace HappyTravel.Edo.Api.Services.Management
...
+                .Bind(DeactivateCounterpartyIfNeed);
```
mingazhev, line 33: If needed

Api/Services/Management/AgentMovementService.cs
```diff
@@ -40,6 +47,14 @@ async Task<Result> ValidateRequest()
             }
 
 
+            async Task<bool> AgentHasNoBookings()
+            {
+                var firstBooking = await _edoContext.Bookings.FirstOrDefaultAsync(b => b.AgentId == agentId && b.AgencyId == sourceAgencyId);
```
mingazhev, line 52: AnyAsync will be more short and clear there

Api/Services/Management/AgentMovementService.cs
AlexeyShok: Fixed

Api/Services/Management/AgentMovementService.cs
AlexeyShok: Fixed

Api/Services/Management/AgentMovementService.cs
AlexeyShok: Fixed

mingazhev (approved):

PR #1002, opened 2021-09-26
added agentId, agencyId to all frontend requests loggings

DFSko:
resolves happy-travel/agent-app-project#633


AlexeyShok (approved):

mingazhev (commented):
I'd suggest to rethink this to have less code duplication and to not remember to add logging manually each time. Maybe middlware or filter to do this, may be using existing

DFSko:
Implemented as middleware in another branch

PR #635, opened 2021-04-19
Feature/add full booking report

DFSko:

PR #1003, opened 2021-09-26
removed getting mapper endpoint from vault

DFSko:
resolves happy-travel/agent-app-project#635


mingazhev (approved):
Don't forget to notify devops to remove redundant value from vault

kirillta (approved):

PR #1005, opened 2021-09-27
added checking status endpoint

DFSko:


Api/Controllers/AgentControllers/PaymentsController.cs
```diff
@@ -132,6 +132,20 @@ public async Task<IActionResult> PayByNGenius(string referenceCode)
...
+        
+        
+        /// <summary>
+        ///     Checks payment status in NGenius
+        /// </summary>
+        /// <param name="referenceCode">Booking reference code</param>
+        [HttpGet("accommodations/bookings/{referenceCode}/pay/status")]
+        [ProducesResponseType(typeof(PaymentStatuses), (int) HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(ProblemDetails), (int) HttpStatusCode.BadRequest)]
+        [AllowAnonymous]
```
mingazhev, line 144: This should not be anonymous

Api/Controllers/AgentControllers/PaymentsController.cs
```diff
@@ -132,6 +132,20 @@ public async Task<IActionResult> PayByNGenius(string referenceCode)
...
+        public async Task<IActionResult> CheckStatusInNGenius(string referenceCode)
```
mingazhev, line 145: Looks more like refreshing status, so need to rename the method and use POST verb (update is occured)

Api/Services/Payments/NGenius/NGeniusClient.cs
```diff
@@ -82,6 +82,20 @@ public async Task<Result> RefundMoney(string paymentId, string orderReference, s
         }
 
 
+        public async Task<Result<PaymentStatuses>> CheckStatus(string orderReference)
```
mingazhev, line 85: GetStatus maybe?

Api/Services/Payments/NGenius/NGeniusPaymentService.cs
```diff
@@ -120,6 +122,29 @@ public async Task<Result<CreditCardRefundResult>> Refund(string paymentId, strin
...
+            if (payment is null)
+                return Result.Failure<PaymentStatuses>($"Payment for {referenceCode} not found");
+
+            var data = JsonConvert.DeserializeObject<CreditCardPaymentInfo>(payment.Data);
+            var (_, isFailure, status) = await _client.CheckStatus(data.ExternalId);
+
+            if (isFailure)
+                return Result.Failure<PaymentStatuses>("Status checking failed");
+
+            payment.Status = status;
```
mingazhev, line 140: Should not we return and do nothing if the status is the same?

Api/Controllers/AgentControllers/PaymentsController.cs
DFSko: Fixed

Api/Controllers/AgentControllers/PaymentsController.cs
DFSko: Fixed

Api/Services/Payments/NGenius/NGeniusClient.cs
DFSko: Fixed

Api/Services/Payments/NGenius/NGeniusPaymentService.cs
DFSko: Fixed

mingazhev (approved):

PR #1017, opened 2021-09-28
Add destinationScope to counterparty contracts

konstyefremov:
Kind of a draft for Destination Scopes realization. Feedback is needed.


Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -297,6 +307,28 @@ private Task<Result> ValidatePolicy(MarkupPolicyData policyData, MarkupPolicy so
...
+                    DestinationMarkupScopeTypes.Accommodation => scopeId.Contains("Accommodation"),
+                    _ => scopeId == ""
+                };
+            }
+
+
+            async Task<bool> DestinationScopeIdExists()
+            {
+                var scopeId = policyData.Settings.DestinationScopeId;
+                var (isSuccess, _, value) = await _mapperClient.GetMappings(new List<string> { scopeId }, "en");
```
konstyefremov, line 327: This loads all the mappings from the mapper. If a country is given it will load all 10-20-30k accommodations in a country.

Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -297,6 +307,28 @@ private Task<Result> ValidatePolicy(MarkupPolicyData policyData, MarkupPolicy so
...
+                return isSuccess && value.Any();
```
konstyefremov, line 328: If the mapper client will return some error it will be not detected

Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -297,6 +307,28 @@ private Task<Result> ValidatePolicy(MarkupPolicyData policyData, MarkupPolicy so
             bool TargetIsValid() => policyData.Target != MarkupPolicyTarget.NotSpecified;
 
 
+            bool DestinationScopeIdIsMatchedWithType()
+            {
+                var scopeId = policyData.Settings.DestinationScopeId;
+                var scopeType = policyData.Settings.DestinationScopeType;
+                return scopeType switch
+                {
```
kirillta, line 315: Markups are time consuming, we haven't time to check containments

Api/Models/Markups/MarkupPolicySettings.cs
```diff
@@ -42,5 +52,16 @@ namespace HappyTravel.Edo.Api.Models.Markups
...
+        /// <summary>
+        ///     Type of destination markup scope
+        /// </summary>
+        public DestinationMarkupScopeTypes DestinationScopeType { get; }
+        
+        
+        /// <summary>
+        ///     Id of scope for destination. Could be Locality_XXXX, City_XXXX, Country_XXXX
+        /// </summary>
+        public string DestinationScopeId { get; }
```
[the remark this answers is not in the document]
konstyefremov: This field stores htId from the mapper. Like Locality_XXXX, where XXXX is some number. 
We can store only XXXX as a number and interfere htId from DestinationScope.
Though, I'm not sure if it's a correct idea. Needs to be discussed.

Api/Services/Markups/AdminMarkupPolicyManager.cs
konstyefremov: This validation occurs when we add or edit markups. When markups are applied to accommodation, it will not be validated.

Api/Models/Markups/MarkupPolicySettings.cs
```diff
@@ -7,14 +8,23 @@ namespace HappyTravel.Edo.Api.Models.Markups
     public readonly struct MarkupPolicySettings
     {
         [JsonConstructor]
-        public MarkupPolicySettings(string description, int templateId,
-            IDictionary<string, decimal> templateSettings, int order, Currencies currency)
+        public MarkupPolicySettings(
+            string description, 
+            int templateId,
+            IDictionary<string, decimal> templateSettings,
```
mingazhev, line 14: We don't place each parameter on the new line, please place them one by one, moving to the new line only after line became too long

Api/Models/Markups/MarkupPolicySettings.cs
```diff
@@ -7,14 +8,23 @@ namespace HappyTravel.Edo.Api.Models.Markups
...
         [JsonConstructor]
-        public MarkupPolicySettings(string description, int templateId,
-            IDictionary<string, decimal> templateSettings, int order, Currencies currency)
+        public MarkupPolicySettings(
+            string description, 
+            int templateId,
+            IDictionary<string, decimal> templateSettings, 
+            int order, 
+            Currencies currency,
+            DestinationMarkupScopeTypes destinationScopeType = DestinationMarkupScopeTypes.NotSpecified,
```
mingazhev, line 17: We don't need to pass the type there, I supposed that we get it from mapper

Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -297,6 +307,28 @@ private Task<Result> ValidatePolicy(MarkupPolicyData policyData, MarkupPolicy so
...
+                    DestinationMarkupScopeTypes.City => scopeId.Contains("Locality"),
```
mingazhev, line 316: This is not the responsibiliity of EDO to validate such scopeIds and location types. Such values should be returned from mapper

Api/Services/Markups/AdminMarkupPolicyManager.cs
mingazhev: This should be written in TODO-comments, not to PR comments. Better if with the link to the task to fix that

mingazhev (approved):

PR #1020, opened 2021-09-30
clean and refactor ngenius service

DFSko:
resolves happy-travel/agent-app-project#662

### DIFF
--- a/Api/Extensions/AgentAgencyExtensions.cs
+++ b/Api/Extensions/AgentAgencyExtensions.cs
@@ -1,0 +1,19 @@
+﻿using HappyTravel.Edo.Api.Models.Agencies;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Api.Models.Payments.NGenius;
+
+namespace HappyTravel.Edo.Api.Extensions
+{
+    public static class AgentAgencyExtensions
+    {
+        public static NGeniusBillingAddress ToBillingAddress(this (AgentContext Agent, SlimAgencyInfo Agency) agentAgencyTuple) 
+            => new ()
+            {
+                FirstName = agentAgencyTuple.Agent.FirstName,
+                LastName = agentAgencyTuple.Agent.LastName,
+                Address1 = agentAgencyTuple.Agency.Address,
+                City = agentAgencyTuple.Agency.City,
+                CountryCode = agentAgencyTuple.Agency.CountryCode
+            };
+    }
+}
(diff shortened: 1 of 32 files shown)


Api/Extensions/AgentAgencyExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using HappyTravel.Edo.Api.Models.Agencies;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Api.Models.Payments.NGenius;
+
+namespace HappyTravel.Edo.Api.Extensions
```
mingazhev, line 5: Please don't place such extensions to this common folder. Extension is used in one or two place, move it to namespaces nearer to usage, may be to the same folder with service where it is used

Api/Extensions/MoneyAmountExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿using HappyTravel.Edo.Api.Models.Payments.NGenius;
+using HappyTravel.Money.Extensions;
+using HappyTravel.Money.Models;
+
+namespace HappyTravel.Edo.Api.Extensions
+{
+    public static class MoneyAmountExtensions
```
mingazhev, line 7: Same there, nearer to NGenius services

Api/Services/Payments/PaymentService.cs
```diff
@@ -0,0 +1,83 @@
...
+using HappyTravel.Edo.Common.Enums;
+using HappyTravel.Edo.Data;
+using HappyTravel.Edo.Data.Payments;
+using HappyTravel.Money.Models;
+using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
+
+namespace HappyTravel.Edo.Api.Services.Payments
+{
+    public class PaymentService : IPaymentService
```
mingazhev, line 16: PaymentsManagementService maybe?
Or PaymentRecordsService?

`PaymentService` sounds like the God-service responsible for all the payments

mingazhev: And as I see it has credit-card specific things. This should be reflected in the name, if this service is used for credit card payments management

Api/Extensions/AgentAgencyExtensions.cs
DFSko: Fixed

Api/Extensions/MoneyAmountExtensions.cs
DFSko: Fixed

Api/Services/Payments/PaymentService.cs
DFSko: Fixed

mingazhev (approved):

PR #1025, opened 2021-10-04
added agent logging middleware

DFSko:
resolves happy-travel/agent-app-project#633

### DIFF
--- a/Api/Infrastructure/Middlewares/AgentRequestLoggingMiddleware.cs
+++ b/Api/Infrastructure/Middlewares/AgentRequestLoggingMiddleware.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using CSharpFunctionalExtensions;
 using HappyTravel.Edo.Api.Services.Agents;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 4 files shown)


Api/Infrastructure/Middlewares/AgentRequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+
+        public async Task InvokeAsync(HttpContext context)
+        {
+            var scopedData = new Dictionary<string, object>();
+
+            try
+            {
+                var agentContext = await _agentContextService.GetAgent();
```
mingazhev, line 26: You can use agentcontextonternal service to avoid exceptions

DFSko: Fixed

Api/Infrastructure/Middlewares/AgentRequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+            _logger = logger;
+        }
+
+
+        public async Task InvokeAsync(HttpContext context)
+        {
+            var scopedData = new Dictionary<string, object>();
+            var (isSuccess, _, agentContext, _) = await _agentContextService.GetAgentInfo();
+
+            if (isSuccess)
```
mingazhev, line 26: I suggest to create scope only if it's needed, when we have agent context

mingazhev (approved):
With a comment

Api/Infrastructure/Middlewares/AgentRequestLoggingMiddleware.cs
DFSko: fixed

PR #1028, opened 2021-10-05
added in-memory markup policy storage

DFSko:
resolves happy-travel/agent-app-project#669


Api/Services/Markups/MarkupPolicyStorage.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using HappyTravel.Edo.Data.Markup;
+
+namespace HappyTravel.Edo.Api.Services.Markups
+{
+    public class MarkupPolicyStorage : IMarkupPolicyStorage
+    {
+        public List<MarkupPolicy> Get(Func<MarkupPolicy, bool> predicate) 
+            => _storage.Where(predicate).ToList();
```
mingazhev, line 11: I think we should wait until markups will be filled there or throw an exception.
We must guarantee that nobody can search with invalid (empty) list of markups

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public class MarkupPolicyStorageUpdater : IHostedService
+    {
+        public MarkupPolicyStorageUpdater(IServiceScopeFactory scopeFactory) 
+            => _scopeFactory = scopeFactory;
+
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
```
mingazhev, line 23: Need to try and catch to not fail when failure occured once. And write exception to sentry

mingazhev: And requested cancellation should not fail as exception too.
Additionally need to check that on graceful shutdown service is stopped normally and cancellation token is valid

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                using (var scope = _scopeFactory.CreateScope())
+                {
+                    var logger = scope.ServiceProvider.GetRequiredService<ILogger<MarkupPolicyStorageUpdater>>();
```
mingazhev, line 27: Why not injecting this logger into constructor?

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
```diff
@@ -0,0 +1,65 @@
...
+                        .OrderByDescending(p => p.Modified)
+                        .Select(p => p.Modified)
+                        .FirstOrDefaultAsync(cancellationToken);
+
+                    if (lastModifiedDate > _lastCheckingDate)
+                    {
+                        var markupPolicies = await context.MarkupPolicies
+                            .ToListAsync(cancellationToken);
+
+                        var markupPolicyStorage = scope.ServiceProvider.GetRequiredService<IMarkupPolicyStorage>();
```
mingazhev, line 41: Why not injecting this in constructor? It is singleton also

DFSko: Fixed

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
DFSko: Fixed

Api/Services/Markups/MarkupPolicyStorage.cs
```diff
@@ -0,0 +1,41 @@
...
+
+
+        public void Set(List<MarkupPolicy> markupPolicies)
+        {
+            _storage = markupPolicies;
+            _isFilled = true;
+        }
+
+
+        private volatile List<MarkupPolicy> _storage = new();
```
kirillta, line 35: Why not a cache instance?

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
```diff
@@ -0,0 +1,80 @@
...
+            => Task.CompletedTask;
+
+
+        private async Task UpdateStorage(CancellationToken cancellationToken)
+        {
+            using var scope = _scopeFactory.CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<EdoContext>();
+            var dateTimeProvider = scope.ServiceProvider.GetRequiredService<IDateTimeProvider>();
+
+            var lastModifiedDate = await context.MarkupPolicies
```
kirillta, line 53: If we know a policy modification date, we may keep them in-memory arbitrary forever

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
```diff
@@ -0,0 +1,80 @@
...
+using HappyTravel.Edo.Api.Infrastructure.Logging;
+using HappyTravel.Edo.Data;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace HappyTravel.Edo.Api.Services.Markups
+{
+    public class MarkupPolicyStorageUpdater : IHostedService
```
kirillta, line 15: I'm not a fan of an idea of hosting a service alongside with an API in the very same solution.

Api/Services/Markups/MarkupPolicyStorage.cs
mingazhev: Why need cache instance?

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
mingazhev: From where should we get this modification date?

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
mingazhev: We can't use external jobs there because we need instance-level memory storage.
Filling cache on demand will slower down some requests, which we'd like to avoid.
Maybe any other idea how to implement?

Api/Services/Markups/MarkupPolicyStorage.cs
kirillta: Because there are no need to re-invent entity life cycles

Api/Services/Markups/MarkupPolicyStorageUpdater.cs
kirillta: It's one line below:
`OrderByDescending(p => p.Modified)`

PR #1029, opened 2021-10-05
Create new master or grant old master with permissions after agent transfer

Raibo:
https://github.com/happy-travel/agent-app-project/issues/552


Api/Services/Management/AgentMovementService.cs
```diff
@@ -32,7 +33,8 @@ public async Task<Result> Move(int agentId, int sourceAgencyId, int targetAgency
                 .Check(_ => UpdateAgencyRelation())
                 .Check(_ => WriteLog())
                 .Check(_ => DeactivateAgencyIfNeeded())
-                .Check(DeactivateCounterpartyIfNeeded);
+                .Check(DeactivateCounterpartyIfNeeded)
```
mingazhev, line 36: May we use `CheckIf` there?

PR #1038, opened 2021-10-07
added ngenius refund support

DFSko:
resolves happy-travel/agent-app-project#664

Not tested because extremely hard to find accommodations with not 100% cancallation penalty


DFSko:
Successfully tested with 70% cancellation penalty
Refund added to queue
image

PR #1072, opened 2021-10-15
moved agencyTreeIds to MarkupSubjectInfo.cs

DFSko:


Api/Services/Markups/Abstractions/MarkupSubjectInfo.cs
```diff
@@ -7,5 +9,6 @@ namespace HappyTravel.Edo.Api.Services.Markups.Abstractions
         public int CounterpartyId { get; init;}
         public int AgencyId { get; init;}
         public int AgentId { get; init;}
+        public List<int> AgencyTreeIds { get; init; }
```
mingazhev, line 12: Parent agency tree ids?

mingazhev: Or upper, something meaningful about what agencies

mingazhev (approved):
With a small comment

PR #1091, opened 2021-10-18
Add destination markups calculation and modify tests

konstyefremov:


Api/Services/Markups/MarkupPolicyService.cs
```diff
@@ -25,14 +25,16 @@ public List<MarkupPolicy> Get(MarkupSubjectInfo subjectInfo, MarkupObjectInfo ob
             agencyTreeIds.Add(agencyId);
             
             var policies = _markupPolicyStorage.Get(p =>
```
mingazhev, line 27: Need more readable approach there

PR #1094, opened 2021-10-20
Feature/724 agency verification

Raibo:
https://github.com/happy-travel/agent-app-project/issues/724

Note that we will have to change notifications, but only when we completely move to agency verification.

We will also need to move all current verification data from `Counterparties` to `Agencies`, also later. 

Old `Counterparty` verification endpoints:
```
api/{v:apiVersion}/admin/counterparties/{counterpartyId}/verify-read-only
api/{v:apiVersion}/admin/counterparties/{counterpartyId}/verify-full-access
api/{v:apiVersion}/admin/counterparties/{counterpartyId}/decline-verification
```

New `Agency` verification endpoints:
```
api/{v:apiVersion}/admin/agencies/{agencyId}/verify-read-only
api/{v:apiVersion}/admin/agencies/{agencyId}/verify-full-access
api/{v:apiVersion}/admin/agencies/{agencyId}/decline-verification
```

### DIFF
--- a/Api/AdministratorServices/AgencyVerificationService.cs
+++ b/Api/AdministratorServices/AgencyVerificationService.cs
@@ -1,0 +1,180 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using CSharpFunctionalExtensions;
+using HappyTravel.DataFormatters;
+using HappyTravel.Edo.Api.Infrastructure;
+using HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions;
+using HappyTravel.Edo.Api.Infrastructure.Options;
+using HappyTravel.Edo.Api.Models.Agents;
+using HappyTravel.Edo.Api.Models.Mailing;
+using HappyTravel.Edo.Api.Models.Management.AuditEvents;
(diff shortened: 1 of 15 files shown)


Api/AdministratorServices/AgencyVerificationService.cs
```diff
@@ -0,0 +1,180 @@
...
+                    CounterpartyName = agency.Name,
+                    State = EnumFormatters.FromDescription(state),
+                };
+
+                return await _notificationService.Send(agent: new SlimAgentContext(master.Id, agency.Id),
+                    messageData: messageData,
+                    notificationType: NotificationTypes.CounterpartyVerificationChanged,
+                    email: master.Email,
+                    templateId: _mailingOptions.CounterpartyVerificationChangedTemplateId);
+            }
```
Raibo, line 150: I left this method somewhat unchanged, because we can repurpose the template to inform about agency verification instead of agency verification later.
We will be able to test this with current mail template.

Options, Data and Template can be ranamed when we change the template and move to Agency verification.

This would be easier than first creating a new template and then deleting the old one.

mingazhev (approved):
May be merged after the code freeze

PR #1134, opened 2021-10-31
added booking info

DFSko:


HappyTravel.Edo.DirectApi/Services/BookingInfoService.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class BookingInfoService
+    {
+        public BookingInfoService(IBookingInfoService bookingInfoService)
+        {
+            _bookingInfoService = bookingInfoService;
+        }
+
+
+        public async Task<Result<Booking>> Get(string referenceCode, AgentContext agent)
```
mingazhev, line 19: This should allow to get the booking not only by our reference code, but also by client reference

DFSko: it will be implemented in happy-travel/agent-app-project#802 task

mingazhev (approved):

PR #1142, opened 2021-11-01
Check that markup is all global when adding discount

konstyefremov:


Api/AdministratorServices/AgencyDiscountManagementService.cs
```diff
@@ -76,7 +76,8 @@ async Task<Result> ValidateTargetMarkup()
                 if (targetMarkup is null)
                     return Result.Failure($"Could not find markup policy with id {createDiscountRequest.TargetMarkupId}");
                 
-                if (targetMarkup.SubjectScopeType != SubjectMarkupScopeTypes.Global)
+                if (targetMarkup.SubjectScopeType != SubjectMarkupScopeTypes.Global
```
mingazhev, line 79: Need to place on the one line or add curly brackets

mingazhev (approved):

PR #775, opened 2021-06-29
Check administrator permission against role

konstyefremov:
resolves https://github.com/happy-travel/agent-app-project/issues/336

PR #1145, opened 2021-11-02
Fix agent context retrieval for admin panel

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/775

### DIFF
--- a/Api/Services/Agents/HttpBasedAgentContextService.cs
+++ b/Api/Services/Agents/HttpBasedAgentContextService.cs
@@ -63,12 +63,15 @@ namespace HappyTravel.Edo.Api.Services.Agents
 
         private async Task<AgentContext> GetAgentContext()
         {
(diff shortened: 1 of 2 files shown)


Api/Services/Agents/HttpBasedAgentContextService.cs
```diff
@@ -63,12 +63,15 @@ public async Task RefreshAgentContext()
 
         private async Task<AgentContext> GetAgentContext()
         {
-            return _tokenInfoAccessor.GetClientId() switch
```
AlexeyShok, line 66: in my opinion, using switch is more clear and understandable in this case than several if checks in a row

mingazhev: But looks like switch is not fully suitable, code is changed

mingazhev (approved):

PR #809, opened 2021-07-08
Feature/aa 396 accept and update hotel confirmation code

AlexeyShok:
Added models for hotel confirmation code. Added endpoint to update booking status from hotel booking confirmation page. Added migration for new table HotelConfirmationHistory

PR #1146, opened 2021-11-02
Add ability to set localityHtId for agency

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/809


Api/Controllers/AgentControllers/AgenciesController.cs
```diff
@@ -9,9 +9,11 @@
 using HappyTravel.Edo.Api.Models.Agencies;
 using HappyTravel.Edo.Api.Models.Agents;
 using HappyTravel.Edo.Api.Models.Invitations;
+using HappyTravel.Edo.Api.Models.Locations;
```
AlexeyShok, line 12: Is this using exactly necessary? I see that a new link has been added and no more changes in the class code.

Api/Services/Agents/AgencyService.cs
```diff
@@ -82,10 +87,24 @@ async Task UpdateAgencyRecord()
                 agencyRecord.VatNumber = editAgencyRequest.VatNumber;
                 agencyRecord.PreferredPaymentMethod = editAgencyRequest.PreferredPaymentMethod;
 
+                if (!string.IsNullOrWhiteSpace(editAgencyRequest.LocalityHtId))
+                {
+                    var (_, isFailure, localityInfo, _) = await _localityInfoService.GetLocalityInfo(editAgencyRequest.LocalityHtId);
```
AlexeyShok, line 92: Perhaps it is better to receive an error message from the _localityInfoservice, rather than inventing the cause of the error yourself?

Api/Controllers/AgentControllers/AgenciesController.cs
```diff
@@ -9,9 +9,11 @@
...
 using HappyTravel.Edo.Api.Services;
 using HappyTravel.Edo.Api.Services.Agents;
 using HappyTravel.Edo.Api.Services.Invitations;
+using HappyTravel.Edo.Api.Services.Locations;
```
AlexeyShok, line 16: And here too?

Api/Controllers/AgentControllers/AgenciesController.cs
```diff
@@ -70,8 +72,8 @@ public async Task<IActionResult> GetSelfAgency()
         [InAgencyPermissions(InAgencyPermissions.PermissionManagement)]
         public async Task<IActionResult> EditSelfAgency([FromBody] EditAgencyRequest request)
         {
+
```
AlexeyShok, line 75: Extra empty line

AlexeyShok (changes requested):

Api/Models/Agencies/EditAgencyRequest.cs
```diff
@@ -7,7 +7,7 @@ namespace HappyTravel.Edo.Api.Models.Agencies
     {
         [JsonConstructor]
         public EditAgencyRequest(string address, string billingEmail, string fax, string phone, string postalCode, string website, string vatNumber,
-            PaymentTypes preferredPaymentMethod)
+            PaymentTypes preferredPaymentMethod, string localityHtId = "")
```
mingazhev, line 10: You can use required value there, we'll update frontend shortly

Api/Services/Agents/AgencyService.cs
```diff
@@ -15,10 +17,13 @@ public class AgencyService : IAgencyService
     {
         public AgencyService(IAdminAgencyManagementService adminAgencyManagementService,
             IDateTimeProvider dateTimeProvider,
-            EdoContext edoContext)
+            ILocalityInfoService localityInfoService,
+            EdoContext edoContext
+            )
```
mingazhev, line 22: Bracket on the same line

Api/Services/Agents/AgencyService.cs
mingazhev: I don't think so, here may be technical error that we should not show to frontend

Api/Services/Agents/AgencyService.cs
```diff
@@ -82,10 +87,24 @@ async Task UpdateAgencyRecord()
                 agencyRecord.VatNumber = editAgencyRequest.VatNumber;
                 agencyRecord.PreferredPaymentMethod = editAgencyRequest.PreferredPaymentMethod;
 
+                if (!string.IsNullOrWhiteSpace(editAgencyRequest.LocalityHtId))
```
mingazhev, line 90: You can use the flow where this field is mandatory

mingazhev: We may not need to update the value if localityHt was not changed

mingazhev:
This PR is totally incorrect

PR #1156, opened 2021-11-03
added client reference code validation service

DFSko:
resolves happy-travel/agent-app-project#814


mingazhev (commented):
Will not work in some cases, when two requests go one after another very fast

mingazhev (approved):

PR #1171, opened 2021-11-09
moved booking refresh options to consul

DFSko:


Api/Infrastructure/Options/BookingStatusUpdate.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using HappyTravel.SuppliersCatalog;
+
+namespace HappyTravel.Edo.Api.Infrastructure.Options
+{
+    public class BookingStatusUpdate
```
mingazhev, line 6: BookingStatusUpdateOptions maybe?

mingazhev (approved):

Api/Infrastructure/Options/BookingStatusUpdate.cs
DFSko: fixed

PR #1172, opened 2021-11-09
Endpoint to get a list of all root agencies

Raibo:
https://github.com/happy-travel/agent-app-project/issues/778

### DIFF
--- a/Api/AdministratorServices/AdminAgencyManagementService.cs
+++ b/Api/AdministratorServices/AdminAgencyManagementService.cs
@@ -66,6 +66,15 @@ namespace HappyTravel.Edo.Api.AdministratorServices
         }
 
 
+        public Task<List<AgencyInfo>> GetRootAgencies(string languageCode = LocalizationHelper.DefaultLanguageCode)
+            => (
+                    from a in _context.Agencies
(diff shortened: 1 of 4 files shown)


Api/AdministratorServices/AdminAgencyManagementService.cs
```diff
@@ -66,6 +66,15 @@ select a.ToAgencyInfo(a.ContractKind, ra.VerificationState, ra.Verified, c.Names
         }
 
 
+        public Task<List<AgencyInfo>> GetRootAgencies(string languageCode = LocalizationHelper.DefaultLanguageCode)
+            => (
+                    from a in _context.Agencies
+                    join c in _context.Countries on a.CountryCode equals c.Code
+                    where a.ParentId == null
+                    select a.ToAgencyInfo(a.ContractKind, a.VerificationState, a.Verified, c.Names, languageCode))
```
mingazhev, line 74: Don't we need "IsActive" value there?

Raibo: No, because there is a use-case that an admin would take inactive agency and activate it.
If we filter out inactive agencies, this would be not possible.

PR #1177, opened 2021-11-10
fixed wrong status condition in ngenis webhook proccessing

DFSko:
resolves happy-travel/agent-app-project#791


Api/Services/Payments/NGenius/NGeniusWebhookProcessingService.cs
```diff
@@ -71,7 +71,7 @@ private async Task TryUpdatePayment(string referenceCode, string paymentId, Paym
                 })
                 .SingleOrDefault();
 
-            if (payment is null || status != payment.Status)
+            if (payment is null || status == payment.Status || status == PaymentStatuses.Authorized && payment.Status == PaymentStatuses.Captured)
```
konstyefremov, line 74: it will be easier to read 3 separate ifs than one big with Or's

mingazhev (commented):
Agree with @konstyefremov

mingazhev (approved):

PR #1179, opened 2021-11-10
Added display markup formula to agency info

mingazhev:


Api/AdministratorServices/CounterpartyManagementService.cs
```diff
@@ -87,14 +77,16 @@ where ar.IsActive
...
-                where a.CounterpartyId == counterpartyId
-                join ra in _context.Agencies on a.Ancestors.Any() ? a.Ancestors[0] : a.Id equals ra.Id
-                select a.ToAgencyInfo(a.ContractKind, ra.VerificationState, ra.Verified, c.Names, languageCode))
-            .ToListAsync();
+            => (
+                    from a in _context.Agencies
+                    join c in _context.Countries on a.CountryCode equals c.Code
+                    join cp in _context.Counterparties on a.CounterpartyId equals cp.Id
+                    where a.CounterpartyId == counterpartyId
+                    join ra in _context.Agencies on a.Ancestors.Any() ? a.Ancestors[0] : a.Id equals ra.Id
```
konstyefremov, line 85: Why a.Ancestors[0]? Maybe it should be First() or Single() or has a comment describing why it's [0]?

konstyefremov (approved):

Api/AdministratorServices/CounterpartyManagementService.cs
Raibo: `First()` crushes the EF engine, `[0]` is the only option that actually translates to SQL.

Raibo: As for why it is the first value in the array, it is because `Ancestors` are filled in specific order: first the root agency, then it's child agency, and so on. 
This guarantees that if there are any values in `Ancestors` at all, then the first value would be the root agency's id.

PR #1187, opened 2021-11-14
Add null check for roleIds

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/830


Api/Services/Agents/AgentService.cs
```diff
@@ -206,7 +206,7 @@ private static List<InAgencyPermissions> GetActualPermissions(AgencyVerification
 
         private static InAgencyPermissions GetInAgencyPermissions(int[] roleIds, List<AgentRole> roles)
         {
-            if (roleIds.Length == 0)
+            if (roleIds == null || roleIds.Length == 0)
```
konstyefremov, line 209: roleIds isn't intended to be null. Maybe it's better not to check this and leave the code as it is?

mingazhev (approved):

PR #1206, opened 2021-11-22
fixed bonus materialization errors

DFSko:


Api/Services/Markups/MarkupBonusMaterializationService.cs
```diff
@@ -102,19 +109,22 @@ private async Task<Result> ApplyBonus(MaterializationData data)
             return await applyBonusTask;
 
 
-            Task<Result> ApplyAgentScopeBonus() 
-                => ApplyAgencyBonus(data.PolicyId, data.ReferenceCode, data.AgencyId, data.Amount);
+            Task<Result> ApplyAgentScopeBonus()
+            {
+                var agencyId = data.SubjectScopeId.Split("-").First();
```
mingazhev, line 114: We have `AgentInAgencyId` for this. But first value is an agent. Please check if we need to change something, migration or something else

DFSko: Fixed

mingazhev (approved):

PR #1213, opened 2021-11-24
fixed exception if supplier connector not configured

DFSko:
resolves happy-travel/agent-app-project#884


mingazhev:
Let's close for now, think that problem is not so urgent. Will try to manage in a better way with supplier management service

PR #1220, opened 2021-11-25
Added account manager name and registration date to root-agencies

Raibo:
https://github.com/happy-travel/ueda/issues/142

also removed all the unused fields.

SQL looks like this:

```
      SELECT a."Id", a."Address", a."Ancestors", a."BillingEmail", a."City", a."ContractKind", a."CounterpartyId", a."CountryCode", a."CountryHtId", a."Created", a."Fax", a."IsActive", a."IsContractUploaded", a."LegalAddress", a."LocalityHtId", a."Modified", a."Name", a."ParentId", a."Phone", a."PostalCode", a."PreferredCurrency", a."PreferredPaymentMethod", a."VatNumber", a."VerificationReason", a."VerificationState", a."Verified", a."Website", t1."FirstName", t1."LastName", c."Names"
      FROM "Agencies" AS a
      INNER JOIN "Countries" AS c ON a."CountryCode" = c."Code"
      LEFT JOIN (
          SELECT d."Id", d."AgencyId", d."AgentId", d."CounterpartyId", d."DisplayFormula"
          FROM "DisplayMarkupFormulas" AS d
          WHERE (d."AgentId" IS NULL)
      ) AS t ON a."Id" = t."AgencyId"
      LEFT JOIN (
          SELECT t0."FirstName", t0."LastName", t0."AgencyId"
          FROM (
              SELECT a1."FirstName", a1."LastName", a0."AgencyId", ROW_NUMBER() OVER(PARTITION BY a0."AgencyId" ORDER BY a0."AgentId", a0."AgencyId", a1."Id") AS row
              FROM "AgentAgencyRelations" AS a0
              INNER JOIN "Agents" AS a1 ON a0."AgentId" = a1."Id"
              WHERE a0."IsActive" AND a0."AgentRoleIds" @> ARRAY[@__accountManagerRoleId_0]::integer[]
          ) AS t0
          WHERE t0.row <= 1
      ) AS t1 ON a."Id" = t1."AgencyId"
      WHERE (a."ParentId" IS NULL)
```


Api/AdministratorServices/AdminAgencyManagementService.cs
```diff
@@ -69,14 +67,23 @@ public Task<Result<AgencyInfo>> GetRoot(int agencyId, string languageCode = Loca
...
-                    join c in _context.Countries on a.CountryCode equals c.Code
-                    from markupFormula in _context.DisplayMarkupFormulas.Where(f => f.AgencyId == a.Id && f.AgentId == null).DefaultIfEmpty()  
-                    where a.ParentId == null
-                    select a.ToAgencyInfo(a.ContractKind, a.VerificationState, a.Verified, c.Names, languageCode, markupFormula == null ? string.Empty : markupFormula.DisplayFormula))
+        public async Task<List<AdminViewAgencyInfo>> GetRootAgencies(string languageCode = LocalizationHelper.DefaultLanguageCode)
+        {
+            var accountManagerRoleId = await _context.AgentRoles
+                .Where(a => a.Name == AccountManagerRoleName)
+                .Select(r => r.Id)
+                .SingleOrDefaultAsync();
```
AlexeyShok, line 75: Why is the role of account manager taken from the database and is it checked that the agent has it? Wasn't it possible to just check the rights with a filter by the permishene at the input to the controller? We have already discussed earlier that we should not bind to roles in the code, only to permissions. Since the names of the roles and the list of permissions included in them can change at any time.

AlexeyShok (changes requested):

Api/AdministratorServices/AdminAgencyManagementService.cs
Raibo: So there are several questions in this comment, as I understand it.

1. Why search for an agent with a specific role, not a permission.
Because the task had a clear description: "Name of the Account manager", not "Name of someone who have the ability to do this or that".
This is for the office, not for us. The office want that, we give that to them, unless there is a really good reason not to.

2. Why not send the definition of the "Account manager" to the input of the controller?
Because this will only shift the responsibility to the front-end, which solves nothing.

3. Why use role name, while role name can change?
A role has only 3 aspects to filter with: Name, permission set and Id. You also mentioned that a permission set might change. Id also might differ between dev and prod. So, we don't have a single aspect to the Role that is solid. 
This means that the problem persists regardless of which aspect of the role we choose to filter it with.

The only way to solve this ambiguity for good I can come up with - is to give roles a new column "Slug" with immutable clear identifier and use it instead.

mingazhev (approved):

Api/AdministratorServices/AdminAgencyManagementService.cs
Raibo: Reworked the changes and now sending empty string there, for now.

AlexeyShok (approved):

PR #1221, opened 2021-11-26
Removed preferred payment method from code

mingazhev:


DFSko (approved):

kirillta:
As I understand, we should reject this one?

PR #1229, opened 2021-11-29
added try-catch and fix error

DFSko:
resolves https://github.com/happy-travel/ueda/issues/189


mingazhev (commented):
Try-catches are not our way. Need to add appropriate validation without throwing exceptions.
Exceptions exist in AgentInAgencyId as a fallback, to ensure its proper work. Normally code should not reach them

kirillta:
So, where's the try-catch?

DFSko:
> So, where's the try-catch?

Removed

DFSko:
> Try-catches are not our way. Need to add appropriate validation without throwing exceptions. Exceptions exist in AgentInAgencyId as a fallback, to ensure its proper work. Normally code should not reach them

Fixed

kirillta (approved):

PR #1238, opened 2021-12-02
Feature/aa 902 add admin slim booking model

DFSko:
resolves https://github.com/happy-travel/agent-app-project/issues/904


Api/AdministratorServices/Models/BookingSlim.cs
```diff
@@ -0,0 +1,19 @@
...
+using HappyTravel.Money.Enums;
+
+namespace HappyTravel.Edo.Api.AdministratorServices.Models
+{
+    public class BookingSlim
+    {
+        public int Id { get; set; }
+        public string ReferenceCode { get; set; }
+        public BookingPaymentStatuses PaymentStatus { get; set; }
+        public int AgentId { get; set; }
```
mingazhev, line 12: Consider using agency and agent names here too, and check in check out seems useful too. Needs to be discussed with frontend

mingazhev (approved):

Api/AdministratorServices/Models/BookingSlim.cs
kirillta: @alexander-plekhanov

DFSko: Added

PR #1265, opened 2021-12-20
Feature/aa 788 update locations of markups

konstyefremov:


Api/Services/Markups/BatchProcessing/MarkupsLocationCheckerService.cs
```diff
@@ -0,0 +1,119 @@
...
+        private Task<Result<string>> ProcessMarkupPolicy(MarkupPolicy policy)
+        {
+            return Result.Success()
+                .Map(CheckSubjectLocation)
+                .Map(CheckDestinationLocation)
+                .Map(UpdateDestinationAccommodation)
+                .Bind(CreateResult);
+
+            
+            async Task<BatchOperationResult> CheckSubjectLocation()
```
mingazhev, line 60: Why do we need BatchOperationResult there?
It is needed on the very end only

Api/Services/Markups/BatchProcessing/MarkupsLocationCheckerService.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                if (policy.SubjectScopeType is not (SubjectMarkupScopeTypes.Country or SubjectMarkupScopeTypes.Locality))
```
mingazhev, line 62: This seems like the failure result but it is not

Api/Services/Markups/BatchProcessing/MarkupsLocationCheckerService.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            Task<BatchOperationResult> ProcessMarkupPolicies() 
+                => Combine(markupPolicies.Select(ProcessMarkupPolicy));
+        }
+
+
+        private Task<Result<string>> ProcessMarkupPolicy(MarkupPolicy policy)
+        {
+            return Result.Success()
+                .Map(CheckSubjectLocation)
```
mingazhev, line 54: We pass the markup through the whole pipe, while the real logic is to process the markup depending on it's location types. I think pipe does not fit the task well

Api/Controllers/AgentControllers/InternalMarkupCheckerController.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Get inactive localities in markups
+        /// </summary>
+        /// <returns>Returns list of inactive localities or empty list</returns>
+        [HttpGet("/inactive-localities")]
+        [ProducesResponseType(typeof(List<string>), (int) HttpStatusCode.OK)]
+        public async Task<IActionResult> GetInactiveLocalities() 
+            => Ok(await _markupsUpdaterService.GetInactiveLocalities());
+        
+        
+        [HttpGet("/update-changed-accommodations")]
```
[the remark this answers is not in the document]
mingazhev: Still not resolved

mingazhev (changes requested):
GET method must not modify anything

AlexeyShok:
PR closed since we don't use HtId in markups at the moment

PR #1267, opened 2021-12-21
Remove counterparty entity

Raibo:
https://github.com/happy-travel/agent-app-project/issues/959

Since previously counterparty did not take place in any logic whatsoever, the only next step to take is remove counterparty entity itself and id references to it from other entities.

The PR might look bulky, but the majority of files affected are tests. This changes cannot be moved out to another PR.

Also renamed field in response of endpoint  `api/{v:apiVersion}/agent`
`Counterparties` -> `AgencyRelations`

Counterparty mentions left in event names, which will be left there forever.

Also in permission names. There is currently another PR https://github.com/happy-travel/edo/pull/1263 with changes there, so to avoid conflicts, did not make any changes here. We will see what will left to change after this and other PR merged.

Also in DirectApi in some enum names.


Api/Models/Agents/AgentDescription.cs
```diff
@@ -48,8 +48,8 @@ namespace HappyTravel.Edo.Api.Models.Agents
         public string Position { get; }
 
         /// <summary>
-        ///     List of counterparties, associated with agent.
+        ///     List of agencies, associated with agent.
         /// </summary>
-        public List<AgentAgencyRelationInfo> Counterparties { get; }
+        public List<AgentAgencyRelationInfo> AgencyRelations { get; }
```
mingazhev, line 53: Isn't this field is used on frontend side? Please check

Raibo: It is used on frontend side.
But we still have to change the naming because now it is misleading.

Api/Models/Agents/AgentDescription.cs
```diff
@@ -48,7 +48,7 @@ namespace HappyTravel.Edo.Api.Models.Agents
         public string Position { get; }
 
         /// <summary>
-        ///     List of counterparties, associated with agent.
+        ///     List of agencies, associated with agent.
         /// </summary>
         public List<AgentAgencyRelationInfo> Counterparties { get; }
```
Raibo, line 53: Moved the actual renaming to another PR https://github.com/happy-travel/edo/pull/1272

mingazhev (approved):

PR #1270, opened 2021-12-22
Feature/aa 953 add infrastructure for sending notifications to admins and example of mailing

AlexeyShok:
Added infrastructure for sending notifications to admins and example of mailing for booking finalized, canceled events


Api/NotificationCenter/Services/NotificationService.cs
```diff
@@ -107,6 +110,71 @@ public async Task<Result> Send(DataWithCompanyInfo messageData, NotificationType
...
+            return await GetRecipients(notificationType)
+                .Bind(recipients => GetNotificationOptions(recipients, notificationType))
+                .Bind(BuildSettings)
+                .Bind(AddNotifications);
+
+
+            async Task<Result<Dictionary<int, string>>> GetRecipients(NotificationTypes notificationType)
+            {
+                var roleIds = _context.AdministratorRoles.Where(r => r.NotificationTypes.Contains(notificationType)).Select(r => r.Id);
+                var recipients = new Dictionary<int, string>();
```
kirillta, line 124: Do we need this line?

Api/NotificationCenter/Services/NotificationService.cs
```diff
@@ -107,6 +110,71 @@ public async Task<Result> Send(DataWithCompanyInfo messageData, NotificationType
...
+
+                return recipientsWithSendingSettings;
+            }
+
+
+            async Task<Result> AddNotifications(List<RecipientWithSendingSettings> recipientsWithSendingSettings)
+            {
+                await _internalNotificationService.AddAdminNotifications(messageData, notificationType, recipientsWithSendingSettings);
+                
+                return Result.Success();
```
kirillta, line 173: Why use `Result` if there is only one return case for the method?

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: We need to define the variable before starting the foreach loop

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: Some refactoring

Api/NotificationCenter/Services/NotificationService.cs
kirillta: Can you execute SelectMany on list of dictionaries?

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: Removed Result from method

Api/NotificationCenter/Services/NotificationService.cs
AlexeyShok: Using SelectMany increases resource consumption, since you first need to convert all data into dictionaries, then use SelectMany and Distinct.

kirillta (approved):

Api/NotificationCenter/Services/NotificationOptionsService.cs
```diff
@@ -12,6 +12,8 @@
 using HappyTravel.Edo.Api.Models.Users;
 using HappyTravel.Edo.Api.NotificationCenter.Models;
 using HappyTravel.Edo.Api.Infrastructure.FunctionalExtensions;
+using System;
+using HappyTravel.Edo.Api.NotificationCenter.Infrastructure;
```
Raibo, line 16: I suggest sorting usings, at least let `System` be at the top.

Raibo (approved):
I approve since I only found non-critical stuff.

Api/NotificationCenter/Services/NotificationOptionsService.cs
AlexeyShok: Sorted by names

PR #1273, opened 2021-12-28
Added model to set suppliers, enabled for multiroom booking

mingazhev:


Api/Models/Settings/AgencyAccommodationBookingSettingsInfo.cs
```diff
@@ -36,5 +36,10 @@ namespace HappyTravel.Edo.Api.Models.Settings
         /// Amount of days, on which deadline policies dated will be shifted towards past. Must not be a positive number
         /// </summary>
         public int? CustomDeadlineShift { get; init; }
+
+        /// <summary>
+        /// Suppliers, allowed for multi-room booking
```
oleg-n-k, line 41: Looks like the dictionary is storing not only suppliers allowed for multiple-room booking

Raibo (approved):
Looks like all fine except what Oleg commented.

mingazhev:
Decided not to implement such thing

PR #1001, opened 2021-09-24
Updated EdoContracts to 2.2.1

mingazhev:
Automatic PR, generated by a script

PR #1276, opened 2021-12-30
always return success for booking refresh from agent app

DFSko:
resolves happy-travel/agent-app-project#1001


AlexeyShok (approved):

Api/Services/Accommodations/Bookings/Management/BookingStatusRefreshService.cs
```diff
@@ -190,9 +192,12 @@ private BookingStatusRefreshState GetUpdatedState(Booking booking, BookingStatus
         }
 
 
-        private bool RefreshCondition(BookingStatusRefreshState state, DateTime date)
+        private bool RefreshCondition(BookingStatusRefreshState state, DateTime date, ApiCaller? apiCaller = null)
         {
-            var delay = GetDelay(state);
+            var delay = apiCaller is {Type: ApiCallerTypes.Agent}
```
mingazhev, line 197: `apiCaller.Type is ApiCallerTypes.Agent` looks more readable for me, pattern matching looks overhead

Api/Services/Accommodations/Bookings/Management/AgentBookingManagementService.cs
```diff
@@ -43,10 +43,10 @@ Task<Result> Cancel(Booking booking)
         public async Task<Result> RefreshStatus(int bookingId, AgentContext agent)
         {
             return await GetBooking(bookingId, agent)
-                .Bind(Refresh);
+                .Tap(Refresh);
 
             
-            Task<Result> Refresh(Booking booking) 
+            Task Refresh(Booking booking)
```
mingazhev, line 49: This looks incorrect for me. Status refresh may be failed but we say that it's all ok there. This may lead to confusion - agent thinks that he\she fetched the actual booking status while this may be not true

Api/Services/Accommodations/Bookings/Management/BookingStatusRefreshService.cs
DFSko: Fixed

Api/Services/Accommodations/Bookings/Management/AgentBookingManagementService.cs
DFSko: Fixed

Api/Controllers/AgentControllers/BookingsController.cs
```diff
@@ -158,10 +158,7 @@ public async Task<IActionResult> FinalizeBooking([FromRoute] string referenceCod
         public async Task<IActionResult> RefreshStatus([FromRoute] int bookingId)
         {
             var agent = await _agentContextService.GetAgent();
-            var (_, isFailure, error) = await _bookingManagementService.RefreshStatus(bookingId, agent);
-            if (isFailure)
-                return BadRequest(ProblemDetailsBuilder.Build(error));
-
+            await _bookingManagementService.RefreshStatus(bookingId, agent);
```
mingazhev, line 161: I think that we should not return Ok result to the client if his action failed.
But may be @kirillta has other opinion?

kirillta: If a status is negative it's OK, but if an error occurs it's a Failure

kirillta (commented):
Dummy comment to delist this PR from my requests

DFSko:
not actual

PR #1277, opened 2022-01-04
updating markup storage on event instead of timer

DFSko:


kirillta (commented):
Why do you need a vender-locked multiplexer for this?

DFSko:
> Why do you need a vender-locked multiplexer for this?

Hided implementation in separate service

kirillta (commented):
I think, you should use remove instead of update. That's more resilient and common cache interaction.

kirillta:
Untitled

PR #1310, opened 2022-01-18
added fields to booking slim model

DFSko:
resolves https://github.com/happy-travel/ueda/issues/227


Api/AdministratorServices/BookingService.cs
```diff
@@ -46,7 +47,11 @@ private IQueryable<BookingSlim> GetBookings(Expression<Func<BookingSlim, bool>>?
                     TotalPrice = booking.TotalPrice,
                     CheckInDate = booking.CheckInDate.DateTime,
                     CheckOutDate = booking.CheckOutDate.DateTime,
-                    Status = booking.Status
+                    DeadlineDate = booking.DeadlineDate != null
+                        ? booking.DeadlineDate.Value.DateTime
+                        : null,
+                    Status = booking.Status,
+                    Supplier = booking.Supplier
```
mingazhev, line 54: Isn't the supplier visible to agents?

mingazhev (approved):

Api/AdministratorServices/BookingService.cs
DFSko: this model used only in administrator endpoints

PR #1332, opened 2022-01-24
Feature/aa 1095 add implementation transfer cancellation policy description to agent app

AlexeyShok:
Added remark to cancellation policy. It is necessary to clarify how to deal with remarks when merging cancellation policies.

### DIFF
--- a/Api/Services/Accommodations/DeadlineMerger.cs
+++ b/Api/Services/Accommodations/DeadlineMerger.cs
@@ -47,7 +47,7 @@ namespace HappyTravel.Edo.Api.Services.Accommodations
                             .FirstOrDefault()
                     );
 
-                    return new CancellationPolicy(date, CalculatePercent(amount));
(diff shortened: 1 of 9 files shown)


Api/Services/Accommodations/DeadlineMerger.cs
```diff
@@ -47,7 +47,7 @@ public static Deadline CalculateMergedDeadline(List<RoomContract> roomContracts)
                             .FirstOrDefault()
                     );
 
-                    return new CancellationPolicy(date, CalculatePercent(amount));
+                    return new CancellationPolicy(date, CalculatePercent(amount), null);    // TODO: Need clarify
```
mingazhev, line 50: This is "simulated" cancellation policy and it cannot have valid description. You can remove TODO there

mingazhev (approved):
With a small comment, TODO - to delete

Api/Services/Accommodations/DeadlineMerger.cs
AlexeyShok: Comment removed

PR #1337, opened 2022-01-27
enabled nullable warnings

DFSko:


HappyTravel.Edo.Notifications/HappyTravel.Edo.Notifications.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
```
mingazhev, line 3: Let's enable this one by one. Starting from "Notifications" only. It is not such a big project and should be easy

DFSko: Fixed

mingazhev (approved):

PR #1339, opened 2022-01-27
Add http clients for supplier options provider

konstyefremov:
https://github.com/happy-travel/agent-app-project/issues/1083

### DIFF
--- a/Api/Infrastructure/Constants/HttpClientNames.cs
+++ b/Api/Infrastructure/Constants/HttpClientNames.cs
@@ -18,5 +18,7 @@
         public const string DacIdentityClient = "DirectApiClientsIdentityClient";
         public const string UsersManagementIdentityClient = "UsersManagementIdentityClient";
         public const string UsersEditIdentityClient = "UsersEditIdentityClient";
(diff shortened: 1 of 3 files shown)


Api/appsettings.json
```diff
@@ -141,5 +141,8 @@
   },
   "DiscountStorageOptions": {
     "Timeout": "0.00:00:10"
+  },
+  "SupplierOptionsProvider": {
+    "Endpoint": "edo/supplier-options-provider"
```
DFSko, line 146: i guess endpoint can be stored in consul

Api/Infrastructure/Constants/HttpClientNames.cs
```diff
@@ -18,5 +18,7 @@ public static class HttpClientNames
         public const string DacIdentityClient = "DirectApiClientsIdentityClient";
         public const string UsersManagementIdentityClient = "UsersManagementIdentityClient";
         public const string UsersEditIdentityClient = "UsersEditIdentityClient";
+        public const string SupplierOptionsProvider = "SupplierOptionsProvider";
```
mingazhev, line 21: Why a separate client is needed? Why not using common edo client?

PR #1347, opened 2022-02-02
removed credit card endpoints and structures

DFSko:
resolves happy-travel/agent-app-project#1116


Api/Models/Payments/NewCreditCardPaymentRequest.cs
```diff
@@ -11,12 +10,10 @@ namespace HappyTravel.Edo.Api.Models.Payments
     {
         [JsonConstructor]
         public NewCreditCardPaymentRequest(string token,
-            string referenceCode, CreditCardInfo cardInfo, bool isSaveCardNeeded)
+            string referenceCode)
```
mingazhev, line 13: Looks suspicious. Does not it break the payfort?
And should not it be renamed too?

mingazhev (approved):
Nice to remove a code, wonderful feeling.
Approved with small comments

Api/Models/Payments/NewCreditCardPaymentRequest.cs
DFSko: Renamed

PR #1370, opened 2022-02-04
implemented grpc connector client

DFSko:


Api/Services/Connectors/GrpcSupplierConnector.cs
```diff
@@ -3,41 +3,162 @@
...
 using HappyTravel.EdoContracts.Accommodations;
+using HappyTravel.EdoContracts.Grpc.Models;
+using HappyTravel.EdoContracts.Grpc.Services;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Prometheus;
 
 namespace HappyTravel.Edo.Api.Services.Connectors;
 
 public class GrpcSupplierConnector : ISupplierConnector
```
mingazhev, line 17: SupplierGrpcConnector

mingazhev (approved):
With a small comment

Api/Services/Connectors/GrpcSupplierConnector.cs
DFSko: Renamed

PR #1395, opened 2022-02-11
Updated images

j-rigg:
Hi @kirillta and @mingazhev,

I'm submitting this so you can look at changes to one of the images while I keep working on language.

The second updated image is the schematic for the search results polling loop. For the original, it looks like no results are returned when `isComplete` is `False`. Based on the docs, it looks like the response actually includes partial results. I updated this. Let me know what you think.

The first image is just cropped, and I fixed spelling in the third image. Nothing exciting.

Thanks,

Jackson


kirillta:
Yes, you're right, a consumer still have an option to use partial results

kirillta (approved):

PR #1442, opened 2022-02-22
FinalPrice renamed to TotalPrice

kirillta:


kirillta:
Merger this one. because it's just a minor wording fix

PR #1470, opened 2022-02-28
Replaced supplier id with supplier code on search steps

mingazhev:


AlexeyShok (approved):

Api/Services/Connectors/ISupplierConnectorManager.cs
```diff
@@ -1,13 +1,7 @@
...
-using System;
-
 namespace HappyTravel.Edo.Api.Services.Connectors
 {
     public interface ISupplierConnectorManager
     {
-        [Obsolete("Integer supplier Id will be removed soon. Use GetByCode method instead")]
-        ISupplierConnector Get(int key);
-
         ISupplierConnector GetByCode(string supplierCode);
```
kirillta, line 5: Could be just `Get` since that's the only method

PR #1471, opened 2022-02-28
Removed unused keys from appsettings

AlexeyShok:
Removed unused keys from appsettings


mingazhev (approved):
Need to remove in consul too

PR #1224, opened 2021-11-26
Technical/aa 733 double booking protection

mingazhev:
https://github.com/happy-travel/agent-app-project/issues/733

The main class is idemponent function executor.
Not well tested for now

PR #1490, opened 2022-03-02
Changed global markups controller to use only a single markup

mingazhev:
https://github.com/happy-travel/agent-app-project/issues/1224

### DIFF
--- a/Api/Controllers/AdministratorControllers/GlobalMarkupsController.cs
+++ b/Api/Controllers/AdministratorControllers/GlobalMarkupsController.cs
@@ -2,7 +2,7 @@ using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
 using HappyTravel.Edo.Api.Filters.Authorization.AdministratorFilters;
 using HappyTravel.Edo.Api.Infrastructure;
-using HappyTravel.Edo.Api.Models.Markups;
+using HappyTravel.Edo.Api.Models.Markups.Global;
 using HappyTravel.Edo.Api.Services.Markups;
 using HappyTravel.Edo.Common.Enums.Administrators;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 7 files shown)


Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -488,6 +338,156 @@ Task WriteAgencyLog(MarkupPolicyEventType eventType)
...
+            async Task<Result<MarkupPolicy>> UpdatePolicy()
+            {
+                policy.Description = settings.Description;
+                policy.TemplateId = settings.TemplateId;
+                policy.TemplateSettings = settings.TemplateSettings;
+                MarkupPolicyValueUpdater.FillValuesFromTemplateSettings(policy, settings.TemplateSettings);
+                policy.Currency = settings.Currency;
+                policy.Modified = _dateTimeProvider.UtcNow();
+                policy.SubjectScopeId = settings.LocationScopeId;
+                // No SubjectScopeType here because changing its type is not allowed
```
konstyefremov, line 474: Maybe instead of making this comment, it is worth adding the check that subjectScopeType remains the same?

Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -488,6 +338,156 @@ Task WriteAgencyLog(MarkupPolicyEventType eventType)
...
+                await _context.SaveChangesAsync();
+                return policy;
+            }
+        }
+
+
+        private async Task<Result> Remove(int policyId)
+        {
+            var (_, isFailure, markupPolicy, error) = await GetPolicy()
+                .Ensure(HasNoDiscounts, "Markup policy has bound discounts")
```
konstyefremov, line 390: Feels like this error message could be better phrased.
"This markup policy has linked discounts"

Api/Controllers/AdministratorControllers/GlobalMarkupsController.cs
```diff
@@ -13,7 +12,7 @@ namespace HappyTravel.Edo.Api.Controllers.AdministratorControllers
 {
     [ApiController]
     [ApiVersion("1.0")]
-    [Route("api/{v:apiVersion}/admin/global-markups")]
+    [Route("api/{v:apiVersion}/admin/global-markup")]
     [Produces("application/json")]
     public class GlobalMarkupsController : BaseController
```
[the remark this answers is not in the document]
mingazhev: Fixed

Api/Services/Markups/AdminMarkupPolicyManager.cs
mingazhev: Fixed

Api/Services/Markups/AdminMarkupPolicyManager.cs
mingazhev: I don't think so. We'll update the models accordingly to make this logic more clear, a bit later

DFSko (approved):

PR #1571, opened 2022-03-25
Add automated api client generation

konstyefremov:
https://github.com/happy-travel/edo/issues/1530


Api/Services/ApiClients/ApiClientService.cs
```diff
@@ -39,8 +40,41 @@ public async Task<Result<ApiClientInfo>> GetCurrent(AgentContext agent)
                 HasDirectContractsFilter = settings.AdditionalSearchFilters.HasFlag(SearchFilters.DirectContractsOnly)
             };
         }
-        
-        
+
+
+        public async Task<GeneratedApiClient> GenerateApiClient(AgentContext agent)
```
DFSko, line 45: look IApiClientManagementService

DFSko:
I guess we need to add abilitiy to user to enter the password manually

konstyefremov:
> I guess we need to add abilitiy to user to enter the password manually

I think it's a good idea to not make such suggestions in PR's, but to discuss them separately in their own time.

AlexeyShok (approved):

Api/Services/Management/ApiClientManagementService.cs
```diff
@@ -101,8 +102,23 @@ Task Remove(ApiClient client)
             Task<Result> WriteAuditLog(ApiClient _)
                 => _managementAuditService.Write(ManagementEventType.AgentApiClientDelete, new AgentApiClientEventData(agentId, agencyId));
         }
-        
-        
+
+
+        public async Task<Result<ApiClientData>> GenerateApiClient(int agencyId, int agentId)
+        {
+            var name = GenericApiClientName + agencyId;
```
kirillta, line 109: Probably, we need a space here

kirillta (commented):
Do you add an option of changing a password in a separate PR?

konstyefremov:
> Do you add an option of changing a password in a separate PR?

In fact, it's implemented here as it uses Set method. If API client doesn't exist, it will generate a new Api Client. If it does exist, it will generate a new password for this client.

kirillta (approved):

PR #1588, opened 2022-03-30
Add log and set markup to zero when it less than zero

DaniyarAskarov:


Api/Services/Markups/MarkupService.cs
```diff
@@ -1,49 +1,58 @@
...
             var policies = _markupPolicyService.Get(subject, destinationInfo);
-            
+
             // Addition markups are not applied for now
             var percentPolicies = policies
                 .Where(p => p.FunctionType == MarkupFunctionType.Percent)
                 .ToList();
 
             var percentSum = percentPolicies.Sum(p => p.Value);
             if (percentSum < 0)
```
DFSko, line 39: i guess we need to log zero percentage too
we can work for free and don't know about it

Api/Services/Markups/MarkupService.cs
```diff
@@ -68,7 +77,9 @@ public class MarkupService : IMarkupService
 
 
         private readonly ICurrencyRateService _currencyRateService;
-        
+
         private readonly IMarkupPolicyService _markupPolicyService;
+
```
DFSko, line 82: unnecessary line break

Api/Services/Markups/MarkupService.cs
```diff
@@ -68,7 +77,9 @@ public class MarkupService : IMarkupService
 
 
         private readonly ICurrencyRateService _currencyRateService;
-        
+
```
DFSko, line 80: unnecessary line break

Api/Services/Markups/MarkupService.cs
DaniyarAskarov: Make sense

AlexeyShok (approved):

PR #1633, opened 2022-04-15
Feature/aa 1342 logging prices

DFSko:


konstyefremov (approved):

Api/Services/Accommodations/Availability/Steps/WideAvailabilitySearch/WideAvailabilitySearchTask.cs
```diff
@@ -93,6 +97,19 @@ await GetAvailability(connectorRequest, languageCode)
             }
 
 
+            void Publish(EdoContracts.Accommodations.Availability availability)
```
kirillta, line 100: Looks like `Publish` is pretty similar for both tasks. Could them be united somewhere?

PR #1636, opened 2022-04-18
Fix validation and restrictions

DaniyarAskarov:


Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -128,13 +128,23 @@ async Task<Result> AddDestinationPolicy()
...
                     {
+                        var valueValidatorMessage = "Markup policy value must be in range (-100..-0.1) or (0.1..100)";
+
+                        v.RuleFor(s => s.Value)
+                            .NotEqual(0)
+                            .WithMessage(valueValidatorMessage)
+                            .GreaterThanOrEqualTo(-100)
+                            .WithMessage(valueValidatorMessage)
+                            .LessThanOrEqualTo(100)
+                            .WithMessage(valueValidatorMessage);
```
AlexeyShok, line 139: Validation is not correct. Instead of checking for greater than or equal to -100 and less than or equal to 100, you need to check the modulo of the value for greater than or equal to 0.1 and less than or equal to 100.

AlexeyShok (approved):

PR #1657, opened 2022-04-22
Fix validation when market doesn't exist

DaniyarAskarov:
Issue: https://github.com/happy-travel/agent-app-project/issues/1291

### DIFF
--- a/Api/Services/Markups/AdminMarkupPolicyManager.cs
+++ b/Api/Services/Markups/AdminMarkupPolicyManager.cs
@@ -150,8 +150,8 @@ namespace HappyTravel.Edo.Api.Services.Markups
 
                             v.RuleFor(m => m.LocationScopeId)
                                 .NotNull()
-                                .MustAsync(AgencyIsExist()!)
-                                .WithMessage(m => $"Agency with Id {m.LocationScopeId} doesn't exist!")
+                                .MustAsync(AgencyAndMarketAreExist()!)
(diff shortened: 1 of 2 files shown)


Api/Services/Markups/AdminMarkupPolicyManager.cs
```diff
@@ -184,9 +184,10 @@ Result ValidateAddLocation(MarkupPolicySettings settings)
                         await _context.Markets.AnyAsync(m => m.Id.ToString() == marketId, cancelationToken);
 
 
-            System.Func<string, System.Threading.CancellationToken, Task<bool>> AgencyIsExist()
+            System.Func<string, System.Threading.CancellationToken, Task<bool>> AgencyAndMarketAreExist()
```
hasmikharutyunyan, line 187: maybe have two different methods ?

konstyefremov, line 187: I'm not happy about the name of the method. It's an obvious grammatical error here. It's either should be AreAgencyAndMarketExist or AgencyAndMarketExist.

konstyefremov, line 187: I don't like System.Func and System.Threading here. Add 

using System;
using System.Threading; 

into beginning of the file.

hasmikharutyunyan (approved):

PR #1697, opened 2022-05-05
added separate model for cached results

DFSko:
resolves happy-travel/agent-app-project#1413


Api/Models/Accommodations/CachedAccommodationAvailabilityResult.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using MongoDB.Bson;
+
+namespace HappyTravel.Edo.Api.Models.Accommodations;
+
+public record CachedAccommodationAvailabilityResult
+{
+    public ObjectId Id { get; init; }
```
kirillta, line 9: Why models so flat? They require some leveling

kirillta (approved):

PR #1414, opened 2022-02-17
Fill SupplierCode in Bookings and SupplierCodes

konstyefremov:
Resolves https://github.com/happy-travel/agent-app-project/issues/1194

PR #1698, opened 2022-05-05
added request hash to cached results

DFSko:
resolves happy-travel/agent-app-project#1414


Api/Models/Availabilities/AvailabilityRequestExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Text.Json;
+
+namespace HappyTravel.Edo.Api.Models.Availabilities;
+
+public static class AvailabilityRequestExtensions
+{
+    public static string ComputeHash(this AvailabilityRequest request)
+    {
+        var bytes = JsonSerializer.SerializeToUtf8Bytes(request);
```
kirillta, line 10: Can you use existing [extension](https://github.com/happy-travel/edo/blob/master/Api/Infrastructure/HashGenerator.cs)?

Api/Models/Availabilities/AvailabilityRequestExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Text.Json;
+
+namespace HappyTravel.Edo.Api.Models.Availabilities;
```
kirillta, line 4: Wrong path for extensions anyway

DFSko: Fixed

Api/Models/Availabilities/AvailabilityRequestExtensions.cs
DFSko: I guess cryptography is not needed here, it slow

kirillta: Just keep both in one file then

DFSko: Fixed

kirillta (approved):

PR #1723, opened 2022-05-12
Fix possible null errors

konstyefremov:


Api/Infrastructure/SupplierConnectors/ConnectorClient.cs
```diff
@@ -124,11 +124,13 @@ private static StringContent BuildContent<T>(T requestContent)
                     return Result.Failure<TResponse, ProblemDetails>(error);
                 }
 
-                return _serializer.Deserialize<TResponse>(jsonTextReader);
+                return _serializer.Deserialize<TResponse>(jsonTextReader) ?? throw new InvalidOperationException();
```
AlexeyShok, line 127: I don't see how this solves the problem with exceptions.

PR #1430, opened 2022-02-20
API documentation wording fixes

kirillta:

PR #1771, opened 2022-05-26
Fixed ModifyAccount without companyBank

timur-mut:
https://github.com/happy-travel/agent-app-project/issues/1430#issuecomment-1138374024

### DIFF
--- a/Api/Infrastructure/ModelExtensions/CompanyAccountExtensions.cs
+++ b/Api/Infrastructure/ModelExtensions/CompanyAccountExtensions.cs
@@ -32,11 +32,11 @@ public static class CompanyAccountExtensions
             Currency = companyAccountInfo.Currency,
             AccountNumber = companyAccountInfo.AccountNumber,
             Iban = companyAccountInfo.Iban,
-            CompanyBankId = companyAccountInfo.CompanyBank.Id,
-            IntermediaryBankName = companyAccountInfo.IntermediaryBank.BankName,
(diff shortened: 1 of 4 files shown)


Api/Infrastructure/ModelExtensions/CompanyAccountExtensions.cs
```diff
@@ -32,11 +32,11 @@ public static CompanyAccount ToCompanyAccount(this CompanyAccountInfo companyAcc
             Currency = companyAccountInfo.Currency,
             AccountNumber = companyAccountInfo.AccountNumber,
             Iban = companyAccountInfo.Iban,
-            CompanyBankId = companyAccountInfo.CompanyBank.Id,
-            IntermediaryBankName = companyAccountInfo.IntermediaryBank.BankName,
-            IntermediaryBankAccountNumber = companyAccountInfo.IntermediaryBank.AccountNumber,
-            IntermediaryBankSwiftCode = companyAccountInfo.IntermediaryBank.SwiftCode,
-            IntermediaryBankAbaNo = companyAccountInfo.IntermediaryBank.AbaNo,
+            CompanyBankId = companyAccountInfo.CompanyBank?.Id ?? default,
```
AlexeyShok, line 35: This should not be default, the bank ID should always be given. We need to check that it is not 0 in service.

AlexeyShok (approved):

PR #1390, opened 2022-02-09
Language edits to Data types

j-rigg:
Hi  @kirillta and @mingazhev,

I edited the **Data types** section as a preview. If this works for you, I will make similar changes to the rest of the file.

Notes:
- I changed some wording to make the text simpler and more concise.
- I put _Booking data_ in the list at the top with static and dynamic data. This will help readers understand that it's a separate category before they continue.
- I mostly used "you" to speak to the reader directly. For example, _You should_ is simpler than _It is recommended that..._.
- I used italics instead of bold to emphasize text in definitions. Usually, bold text is for headings or UI elements.
- The document mostly uses American English, so I kept that style. I can also convert this to British spelling.

Let me know if you have any questions or suggestions. Thanks!

Jackson

PR #1791, opened 2022-05-30
Get default settings for agency

konstyefremov:
https://github.com/happy-travel/agent-app-project/issues/1390#issuecomment-1137383994


Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -25,15 +26,75 @@ public class AgencySystemSettingsManagementService : IAgencySystemSettingsManage
 
         public async Task<Result<AgencyAccommodationBookingSettings>> GetAvailabilitySearchSettings(int agencyId)
         {
-            return await CheckAgencyExists(agencyId)
-                .Bind(GetSettings);
+            var agency = await _context.Agencies.SingleOrDefaultAsync(a => a.Id == agencyId);
+            if (Equals(agency, default))
```
DFSko, line 30: because agency is reference type i guess you can checking for null

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -25,15 +26,75 @@ public class AgencySystemSettingsManagementService : IAgencySystemSettingsManage
...
+                return Result.Failure<AgencyAccommodationBookingSettings>("No agency exist");
+            
+            if (agency.ContractKind == null)
```
DFSko, line 33: i prefer to use 'is' instead of '=='

konstyefremov: > i prefer to use 'is' instead of '=='

any reason why it's better?

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
```diff
@@ -25,15 +26,75 @@ public class AgencySystemSettingsManagementService : IAgencySystemSettingsManage
...
+            AgencyAccommodationBookingSettings MergeSettings(AgencyAccommodationBookingSettings? rootSettings, AgencyAccommodationBookingSettings? agencySettings)
             {
-                var existingSettings = await _context.AgencySystemSettings.SingleOrDefaultAsync(s => s.AgencyId == agencyId);
+                var isSupplierVisible = false;
+                var isDirectContractVisible = false;
+                var customDeadlineShift = 0;
+                var aprMode = AprMode.Hide;
+                var passedDeadlineOffersMode = PassedDeadlineOffersMode.Hide;
+
+                if (contractKind == ContractKind.VirtualAccountOrCreditCardPayments)
```
AlexeyShok, line 58: I think this method does not need to change the settings. This logic is already implemented in the settings change endpoint

Api/AdministratorServices/AgencySystemSettingsManagementService.cs
DFSko: == in user-defined type can be overloaded and this condition may be broken

AlexeyShok (approved):

PR #1851, opened 2022-06-10
renamed analytics indexes

DFSko:


Api/Infrastructure/Analytics/ElasticAnalyticsService.cs
```diff
@@ -27,7 +27,7 @@ public void LogEvent<TEventData>(in TEventData eventData, string name, in AgentA
             var date = _dateTimeProvider.UtcNow();
             var environmentName = _environment.EnvironmentName.ToLowerInvariant();
 
-            var indexName = $"{environmentName}-{ServicePrefix}-{name}-{date:yyyy-MM-dd}";
+            var indexName = $"{environmentName}-{ServicePrefix}-{name}-{date:yyyy.MM}";
```
AlexeyShok, line 30: Maybe we need a day too?

DFSko: We need one index per month

evm-andrey (approved):

Issue #1941, opened 2022-06-29
Emails delivery service


evm-andrey:
Develop a service for delivering emails using the nats queue.

In edo, when an event is sent to send an email notification, send it to the message queue, where the new service will consumer and process the tasks.

It is also important to store the submission history in the postgres database, similar to how it is implemented in edo.

All accounts to access services in a secure vault

Health checks and metrics (number of jobs processed successfully/failed, in progress, etc.) should be implemented.

PR #1942, opened 2022-06-30
replaced mail sending with messaqe queue

DFSko:
#1941

### DIFF
--- a/Api/Infrastructure/MailSenderWithCompanyInfo.cs
+++ b/Api/Infrastructure/MailSenderWithCompanyInfo.cs
@@ -1,53 +1,39 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using CSharpFunctionalExtensions;
-using FluentValidation;
+using HappyTravel.Edo.Api.Infrastructure.Constants;
 using HappyTravel.Edo.Api.Models.Company;
 using HappyTravel.Edo.Api.Models.Mailing;
+using HappyTravel.Edo.Api.Models.Messaging;
 using HappyTravel.Edo.Api.Services.Company;
-using HappyTravel.MailSender;
+using HappyTravel.Edo.Api.Services.Messaging;
 
 namespace HappyTravel.Edo.Api.Infrastructure
 {
     public class MailSenderWithCompanyInfo
     {
-        public MailSenderWithCompanyInfo(IMailSender mailSender, ICompanyService companyService)
+        public MailSenderWithCompanyInfo(IMessageBus messageBus, ICompanyService companyService)
         {
-            _mailSender = mailSender;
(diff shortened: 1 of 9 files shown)


Api/Models/Messaging/MailMessage.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+
+namespace HappyTravel.Edo.Api.Models.Messaging;
+
+public class MailMessage
```
suhan3z, line 6: Please add comments to public fields

Api/Models/Messaging/MailMessage.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public string TemplateId { get; set; } = string.Empty;
+    public IEnumerable<string> Recipients { get; set; } = Array.Empty<string>();
+    public object? Data { get; set; }
```
suhan3z, line 10: Can it really be empty? why not string?

Api/Infrastructure/MailSenderWithCompanyInfo.cs
```diff
@@ -1,53 +1,39 @@
...
-
-
-        public async Task<Result> Send(string templateId, IEnumerable<string> recipientAddresses, DataWithCompanyInfo messageData)
+        public async Task Send(string templateId, IEnumerable<string> recipientAddresses, DataWithCompanyInfo messageData)
         {
             var (_, isFailure, companyInfo, _) = await _companyService.GetCompanyInfo();
             messageData.CompanyInfo = isFailure ? new CompanyInfo() : companyInfo;
 
-            return await _mailSender.Send(templateId, recipientAddresses, messageData);
+            _messageBus.Publish(MessageBusTopics.SendMail, new MailMessage
```
suhan3z, line 27: Is method sync?

DFSko: > Is method sync?

yes, method just run a task and forget

Api/Models/Messaging/MailMessage.cs
DFSko: we can e-mail template without data

AlexeyShok (approved):

Api/Models/Messaging/MailMessage.cs
DFSko: Fixed

Api/Models/Messaging/MailMessage.cs
suhan3z: The problem with the object is that we have to know the type on the sending service side. We either have to make a generic assembly with these types, or we have to contract on the swagger side and generate these types

Issue #1584, opened 2022-03-30
Search issue


sergeyf-QA:
@AlexeyShok 
I have the agent who always takes blank 1st step search result (perhaps this is after I added location markup for him?).
This issue is reproduced for him only.
You can find the agent **fabirs10** in passbolt.

What is in logs:

`Supplier availability search exception on supplier 'https://etg-dev.happytravel.com/api/1.0/'Exception: 
System.ArgumentException: Invalid markup settings
   at HappyTravel.Edo.Api.Services.Markups.MarkupService.ApplyMarkups[TDetails](MarkupSubjectInfo subject, MarkupDestinationInfo destinationInfo, TDetails details, Func`3 priceProcessFunc, Action`1 logAction) in /src/Api/Services/Markups/MarkupService.cs:line 67
   at HappyTravel.Edo.Api.Services.Accommodations.Availability.PriceProcessor.ApplyMarkups[TDetails](MarkupSubjectInfo subjectInfo, TDetails details, Func`3 priceProcessFunc, Func`2 getMarkupDestinationFunc, Action`1 logAction) in /src/Api/Services/Accommodations/Availability/PriceProcessor.cs:line 42
   at HappyTravel.Edo.Api.Services.Accommodations.Availability.Steps.WideAvailabilitySearch.WideAvailabilityPriceProcessor.ApplyMarkups(List`1 results, AgentContext agent) in /src/Api/Services/Accommodations/Availability/Steps/WideAvailabilitySearch/WideAvailabilityPriceProcessor.cs:line 31
   at CSharpFunctionalExtensions.AsyncResultExtensionsBothOperands.Map[T,K,E](Task`1 resultTask, Func`2 func)
   at CSharpFunctionalExtensions.AsyncResultExtensionsLeftOperand.Map[T,K,E](Task`1 resultTask, Func`2 func)
   at CSharpFunctionalExtensions.AsyncResultExtensionsBothOperands.Tap[T,E](Task`1 resultTask, Func`2 func)
   at CSharpFunctionalExtensions.AsyncResultExtensionsBothOperands.Tap[T,E](Task`1 resultTask, Func`1 func)
   at CSharpFunctionalExtensions.AsyncResultExtensionsLeftOperand.Finally[T,K,E](Task`1 resultTask, Func`2 func)
[...]

PR #2008, opened 2022-07-14
Added accommodation image to S3 backet

timur-mut:
https://github.com/happy-travel/agent-app-project/issues/1584

### DIFF
--- a/Api/Services/Files/BookingPhotoLoadingService.cs
+++ b/Api/Services/Files/BookingPhotoLoadingService.cs
@@ -49,10 +49,10 @@ public class BookingPhotoLoadingService : IBookingPhotoLoadingService
                     return Result.Failure<Booking>($"Could not find booking");
                 
                 if (booking.AccommodationInfo?.Photo?.SourceUrl is null)
-                    return Result.Failure<Booking>($"Booking photo source SourceUrl is null");
+                    return Result.Failure<Booking>($"Booking photo SourceUrl is null");
                 
                 if (booking.AccommodationInfo.Photo.SourceUrl.Contains(key))
-                    return Result.Failure<Booking>($"Booking photo source already loaded to S3");
+                    return Result.Failure<Booking>($"Booking photo has already loaded to S3");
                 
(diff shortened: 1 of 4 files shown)


Api/Services/Files/BookingPhotoLoadingService.cs
```diff
@@ -0,0 +1,68 @@
...
+
+
+    public void StartBookingPhotoLoading(int bookingId)
+    {
+        Task.Run(async () =>
+        {
+            var booking = await _context.Bookings.FindAsync(bookingId);
+            if (booking?.AccommodationInfo?.Photo is null)
+                return;
+
```
AlexeyShok, line 34: Can add a check and exit if the photo is already in our S3 bucket?

Api/Services/Files/BookingPhotoLoadingService.cs
```diff
@@ -0,0 +1,68 @@
...
+            await UploadFile(booking.AccommodationInfo.Photo.SourceUrl, bookingId)
+                .Map(newUrl => booking.AccommodationInfo.Photo.SourceUrl = newUrl)
+                .Tap(_ => _context.Update(booking))
```
AlexeyShok, line 37: It is better to put this and the next line in a separate method

Api/Services/Files/BookingPhotoLoadingService.cs
```diff
@@ -0,0 +1,68 @@
...
+                .Tap(_ => _context.SaveChangesAsync())
+                .OnFailure(message => 
+                    _logger.LogError($"Booking photo loading error, booking id: {bookingId}, message: {message}"));
+        });
+    }
+
+
+    private async Task<Result<string>> UploadFile(string url, long bookingId)
+    {
+        var key = $"{_imageFileServiceOptions.S3FolderName}/{bookingId}/photo.jpg";
```
AlexeyShok, line 47: Would it be better to name accommodation.jpg?

AlexeyShok (approved):

PR #2056, opened 2022-07-29
update grpc connector clients on message bus event

DFSko:
resolves happy-travel/agent-app-project#1501


AlexeyShok (approved):

Api/Services/Connectors/GrpcClientsStorageUpdater.cs
```diff
@@ -0,0 +1,43 @@
...
+
+
+    protected override Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        _connection.SubscribeAsync(MessageBusTopics.MarkupPolicyUpdated, (_, m) =>
+        {
+            _logger.LogSupplierUpdateMessageReceived();
+            
+            var supplier = JsonSerializer.Deserialize<SlimSupplier>(m.Message.Data);
+            if (supplier is null)
```
suhan3z, line 30: Don't we need to log something, in this case?

Api/Services/Connectors/IGrpcClientsStorage.cs
```diff
@@ -0,0 +1,11 @@
+using HappyTravel.EdoContracts.Grpc.Services;
+using HappyTravel.SupplierOptionsClient.Models;
+
+namespace HappyTravel.Edo.Api.Services.Connectors;
+
+public interface IGrpcClientsStorage
+{
+    IConnectorGrpcService Get(SlimSupplier supplier);
```
suhan3z, line 8: Comments

DFSko: comment added

Api/Services/Connectors/GrpcClientsStorageUpdater.cs
DFSko: added logging

PR #2106, opened 2022-08-10
added agent and agency names to analytics events

DFSko:


Api/Services/Analytics/IBookingAnalyticsService.cs
```diff
@@ -1,19 +1,15 @@
...
 namespace HappyTravel.Edo.Api.Services.Analytics
 {
     public interface IBookingAnalyticsService
     {
-        public void LogWideAvailabilitySearch(in AgentContext agent);
-        public void LogAccommodationAvailabilityRequested(in Accommodation accommodation, in AgentContext agent);
-        public void LogBookingOccured(Booking booking, in AgentContext agent);
-        public void LogBookingConfirmed(Booking booking);
-        public void LogBookingCancelled(Booking booking);
+        public void LogWideAvailabilitySearch(in AgentInfo agentInfo);
```
timur-mut, line 9: why do you use `in` parameter modifier? 
It is looking like Log methods would modify `AgentInfo` parameter.

DFSko: > why do you use `in` parameter modifier? It is looking like Log methods would modify `AgentInfo` parameter.

argument passed as reference instead of copy for performance issues

timur-mut: Maybe it is better to use record instead of structure?

DFSko: struct is faster

DaniyarAskarov (approved):

PR #2168, opened 2022-09-09
Feature/aa 1783 account credit limit by currencies

DaniyarAskarov:
Issue: https://github.com/happy-travel/agent-app-project/issues/1783
Not tested yet

1) ContractKind change endpoint's api model was changed to receive CreditLimit with AvailableCurrencies. When ContractKind changes to VirtualAccountOrCreditCardPayments virtual accounts with CreditLimit for all currencies would be created. If account with this currency already created CreditLimits sets to account.
2) FullAccess verification for VirtualAccountOrCreditCardPayments should require CreditLimit and then set this CreditLimit for created or existed accounts.
3) Check if old data in Agency transfered to each AgencyAccount with default currency(USD)
4) Check if notifications about CreditLimit run out send for each account with different currencies if it needed.

Notifications about CreditLimit message data would store new field - Currency. It means that template should be changed @DenisVelinen 
Also because of api models were change @alexander-plekhanov  should be mentioned about that. 

Example of json's object AvailableCurrencies with creditLimit:
For ContractKind = VirtualAccountOrCreditCardPayments: 
{
   "USD" : 20000,
   "AED" : 40000,
   "EUR" : 10000
}

For ContractKind = OfflineOrCreditCardPayments:
{
  "USD" : null,
  "AED" : null,
  "EUR" : null
}


Api/AdministratorServices/AdminAgencyManagementService.cs
```diff
@@ -258,24 +264,24 @@ Task<Result> ValidateChange(ContractKindChangeRequest request)
                     }, request);
 
 
+            async Task SetAvailableCurrencies(Agency agency)
+            {
+                var parentSettings = await _context.AgencySystemSettings.SingleOrDefaultAsync(s => s.AgencyId == agency.ParentId);
+                parentAvailableCurrencies = parentSettings?.AccommodationBookingSettings?.AvailableCurrencies ?? new List<Currencies>();
```
AlexeyShok, line 270: Perhaps if there are no currencies, you need to set the default currency? It's not good that a method sets the value of an external variable instead of returning it.

AlexeyShok (approved):

PR #2239, opened 2022-10-04
Feature/aa 1719 make title an optional property2

AlexeyShok:
Added validation for passengers in booking request


AlexeyShok:
https://github.com/happy-travel/agent-app-project/issues/1719

PR #2370, opened 2022-11-02
Feature/aa 1942 add fields to booking summary report for agent

DaniyarAskarov:
Issue: https://github.com/happy-travel/agent-app-project/issues/1942


DFSko (approved):

Api/Services/Accommodations/Bookings/Mailing/BookingReportsService.cs
```diff
@@ -161,6 +162,13 @@ static BookingSummaryNotificationData.BookingData CreateBookingData(Booking book
                         Status = EnumFormatters.FromDescription(booking.Status),
                         PaymentType = EnumFormatters.FromDescription(booking.PaymentType)
                     };
+
+
+                string GetAgentFormattedName(Booking booking)
+                    => _context.Agents
+                        .Where(a => a.Id == booking.AgentId)
+                        .Select(a => $"{a.FirstName} {a.LastName}")
```
AlexeyShok, line 170: What happens if an agent with this Id is not found?

AlexeyShok (approved):

Api/Models/Mailing/BookingSummaryNotificationData.cs
```diff
@@ -14,6 +14,7 @@ public class BookingSummaryNotificationData : DataWithCompanyInfo
         public class BookingData
```
suhan3z, line 14: Lack of comments

Api/Services/Accommodations/Bookings/Mailing/BookingReportsService.cs
```diff
@@ -147,10 +147,17 @@ async Task<(List<EmailAndSetting>, List<Booking>)> GetBookings(List<EmailAndSett
                 return results;
 
 
-                static BookingSummaryNotificationData.BookingData CreateBookingData(Booking booking)
-                    => new()
+                BookingSummaryNotificationData.BookingData CreateBookingData(Booking booking)
+                {
+                    var agent = _context.Agents
```
suhan3z, line 152: Do we need to show only one agent?

suhan3z (changes requested):
According to constants

Api/Services/Accommodations/Bookings/Mailing/BookingReportsService.cs
DaniyarAskarov: Per each booking has only one agent who booked it. According to task this agent's name should be added to mail template

Api/Models/Mailing/BookingSummaryNotificationData.cs
DaniyarAskarov: This class and mail templates were already created without comments. If we need to apply comments to already existed functional, we need to create separate task to cover the whole application with comments.

evm-andrey: > This class and mail templates were already created without comments. If we need to apply comments to already existed functional, we need to create separate task to cover the whole application with comments.

You can start leaving comments little by little within the framework of the current task.

Api/Services/Accommodations/Bookings/Mailing/BookingReportsService.cs
suhan3z: If it's only one, then why we try to find him through firstOrDefault?

suhan3z (approved):

Api/Services/Accommodations/Bookings/Mailing/BookingReportsService.cs
DaniyarAskarov: I changed it SingleOrDefault. But it still one agent who booked it. I get it from db which contain the whole bunch of agents

PR #2495, opened 2022-12-19
Feature/delete type from agent agency relation

DaniyarAskarov:


Api/AdministratorServices/BalanceNotificationsService.cs
```diff
@@ -79,7 +79,8 @@ public async Task<Result> NotifyCreditLimitRunOutBalance(CancellationToken cance
                 var masterAgents = await (
                     from a in _context.Agents
                     join rel in _context.AgentAgencyRelations on a.Id equals rel.AgentId
-                    where rel.AgencyId == agency.Id && rel.Type == AgentAgencyRelationTypes.Master
+                    join roles in _context.AgentRoles on "Master Agent" equals roles.Name
```
AlexeyShok, line 82: It is not correct to check the presence of the agent role by name. It is necessary to check whether the agent has the appropriate permission

Api/Services/Agents/AgentService.cs
```diff
@@ -62,10 +62,19 @@ public AgentService(EdoContext context, IDateTimeProvider dateTimeProvider, IAge
 
         public async Task<Result<Agent>> GetMasterAgent(int agencyId)
         {
-            var master = await (from a in _context.Agents
-                join rel in _context.AgentAgencyRelations on a.Id equals rel.AgentId
-                where rel.AgencyId == agencyId && rel.Type == AgentAgencyRelationTypes.Master
-                select a).FirstOrDefaultAsync();
+            var masterRole = await _context.AgentRoles
+                .FirstOrDefaultAsync(ar => ar.Name == "Master Agent");
```
AlexeyShok, line 66: The same

AlexeyShok (approved):

